PR #13, opened 2018-01-25
Fix formatting in HttpTrigger sample

elprans:


christopheranderson:
CLA is already signed it looks like (look under "all checks have passed"). So you're all good!

PR #14, opened 2018-02-02
PoC.  HttpTrigger2 example works.

1st1:
This is a WIP PR.  So far only the HttpTrigger2 example works.

Immediate ToDo list, ordered by priority, for next PRs:

1. Implement a Functional unittests harness.  Add a test for
   HttpTrigger2 example.

2. Add support for azurefuncs.HttpResponse and azurefuncs.Context.

3. Add support for azurefuncs.Out.

4. Rewrite Dispatcher in async/await, adapting GRPC
   Future objects into asyncio Future objects.

Just in case, I'm developing on macOS, and my current script to launch the host with Python worker looks like this:

```shell
#!/bin/bash

cd /Users/yury/dev/ms/azure-functions-python-worker/Samples/

env AzureWebJobsScriptRoot=/Users/yury/dev/ms/azure-functions-python-worker/Samples/ workers:config:path=/Users/yury/dev/ms/azure-functions-python-worker/ dotnet /Users/yury/dev/ms/SiteExtensions/Functions/Microsoft.Azure.WebJobs.Script.WebHost.dll
```


pragnagopa:
@1st1 - Can you please update the PR without the generated files and the proto file? we just want to keep the script that generates the grpc files as part of the build script.

1st1:
What script do you guys usually use? We need to have it to test locally on our machines.

Samples/HttpTrigger2/function.json
```diff
@@ -0,0 +1,17 @@
...
+    {
+      "authLevel": "anonymous",
+      "type": "httpTrigger",
+      "direction": "in",
+      "name": "req"
+    },
+    {
+      "type": "http",
+      "direction": "out",
+      "name": "res"
```
asavaritayal: Output binding tied to the return value must use `name : $return`

Samples/HttpTrigger2/main.py
```diff
@@ -0,0 +1,5 @@
+import azurefuncs
```
asavaritayal: To align with all other Azure x Python packages, can we rename the module to `azure.functions` instead? 

Example - 
```python
def main(req: func.HttpRequest):
        return 'Hello World!'
```

Samples/HttpTrigger2/main.py
```diff
@@ -0,0 +1,5 @@
...
+
+
+def main(req: azurefuncs.HttpRequest):
```
asavaritayal: Is the type hint here required? Wondering if we can decipher it using the `type` property in `function.json`.

azureworker/types.py
```diff
@@ -0,0 +1,89 @@
...
+class HttpRequest(azurefuncs.HttpRequest):
+    """An HTTP request object.
+
+    The implemented API is a subset of Flask's Request:
+        http://flask.pocoo.org/docs/0.12/api/#incoming-request-data
+    """
+
+    __slots__ = ('_method', '_url', '_headers', '_params', '_body')
+
+    def __init__(self, method: str, url: str,
```
asavaritayal: We want to expose query string parameters here as well. (You likely already have this on the larger backlog - adding here as an fyi/reminder.)

pragnagopa:
@1st1 - .sh file would be fine for the script.

Samples/HttpTrigger1/function.json
```diff
@@ -3,7 +3,7 @@
   "disabled": false,
   "bindings": [
     {
-      "authLevel": "function",
+      "authLevel": "anonymous",
```
pragnagopa: Nit: Change  authlevel to function as that's the default

Samples/HttpTrigger2/main.py
pragnagopa: Use functions instead for func

python/worker.py
```diff
@@ -0,0 +1,38 @@
...
+import traceback
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Python Azure Functions Worker')
+    parser.add_argument('--host')
+    parser.add_argument('--port', type=int)
+    parser.add_argument('--workerId', dest='worker_id')
+    parser.add_argument('--requestId', dest='request_id')
```
pragnagopa: Do we see a friendly error message if any of the arguments are missing?

azurefuncs/abc.py
```diff
@@ -0,0 +1,58 @@
+import abc
```
pragnagopa: Nit: could we rename this file something other than abc?

pragnagopa:
For the other functions repos we use dev branch for latest changes and master to reflect the released bits. We should follow the same process here. Can you please submit a PR against the dev branch after deleting the generated files?

python/worker.py
1st1: Yes, the argparse module takes care of that.

azurefuncs/abc.py
1st1: `abc` is a standard Python abbreviation for Abstract Base Class and it's a pretty standard naming. Do you have a better name in mind?

Samples/HttpTrigger1/function.json
1st1: Sure, will do

Samples/HttpTrigger2/function.json
1st1: Makes sense.

1st1:
Closing this one. I'll resubmit against the 'dev' branch.

1st1:
> For the other functions repos we use dev branch for latest changes and master to reflect the released bits. We should follow the same process here. Can you please submit a PR against the dev branch after deleting the generated files?

Wait.. I don't see the `dev` branch in this repo. Should we create it? 

cc @pragnagopa @asavaritayal

azurefuncs/abc.py
pragnagopa: Looks like there is a standard library abc. Just wanted to make sure our naming is not confusing.

1st1: Yeah, that could be the case, but I'd prefer people not to import this module from `azure.functions` at all.  I guess we should just rename it to `_abc.py`, so that people will be forced to use `azure.functions` directly (and not `azure.functions.abc`.

PR #18, opened 2018-02-16
Simpler gRPC bindings gen option; script-root param for testutils; faster bind

1st1:


1st1:
cc @asavaritayal @christopheranderson

Issue #19, opened 2018-02-16
Validate and add meta data to loaded functions


1st1:
The `FunctionLoadRequest` message contains information about bindings, including the type of the `$return` binding.  We can use that information to validate the signature of the loaded function, as well as creating a dedicated codec for its return value.

1st1:
This can now be closed, the change has been implemented in the `master` branch.

Issue #11, opened 2018-01-25
Logging


asavaritayal:
In order to write traces during an execution, we'll use Python's **logging** module. The runtime will instantiate ```logger``` as a keyword-only required argument of type **logging.Logger**. 

The module provides various convenience methods that let you write to the console at more than one trace level. 

Method | Description
--------|------------
**debug**(message)|Detailed information, typically of interest only when diagnosing problems.
**info**(message)| Confirmation that things are working as expected.
**warning**(message)| An indication that something unexpected happened, or indicative of some problem in the near future (e.g. ‘disk space low’). The software is still working as expected.
**error**(message)| Due to a more serious problem, the software has not been able to perform some function.
**critical**(message)| A serious error, indicating that the program itself may be unable to continue running.

The logger object will be accessible in two ways: 

1. Via the reserved keyword ```logger```
```python
# main.py
def main(logger):
	logger.warning('Something has happened!')
```
2. By declaring ```Logger``` type for a method argument
```python
# main.py
def main(foo: Logger):
	foo.error('Type mismatch')
```

By default, the effective level for logger will be set to INFO. To modify this level, use the ```setLevel()``` method provided by the logging module. Example - 
```python
# main.py
def main(logger):
	logger.setLevel(logging.ERROR)
        logger.debug('Effective trace level changed to error.') # not logged
        logger.error('Something has happened.') # logged
```

**Note:** Since we use a single Python process per function app, using the inbuilt **print** utility will output to our system logs. These are not available to the user and hard to distinguish on a per function basis.

1st1:
I'm looking at our GRPC schema:

```protobuf
message RpcLog {
  // Matching ILogger semantics
  // https://github.com/aspnet/Logging/blob/9506ccc3f3491488fe88010ef8b9eb64594abf95/src/Microsoft.Extensions.Logging/Logger.cs
  enum Level {
    Trace = 0;
    Debug = 1;
    Information = 2;
    Warning = 3;
    Error = 4;
    Critical = 5;
    None = 6;
  }

  string invocation_id = 1;
  string category = 2;
  Level level = 3;
  string message = 4;
  string event_id = 5;
  RpcException exception = 6;

  // json serialized property bag, or could use a type scheme like map<string, TypedData>
  string properties = 7;
}
```

And it seems like Azure Functions logging is a bit different from the standard Python `logging` module:

* Python doesn't have a `Trace` debug level.
[...]

PR #21, opened 2018-02-19
Add support for logging; "out" bindings; new testing infra

1st1:
Summary:

1. Add support for logging. Anything that gets logged via the Python `logging` module is automatically communicated to WebHost via `RpcLog` messages.  `invocation_id` is preserved when a logging event occurs during a function invocation.

2. Add rudimentary support for "out" bindings.

3. Implement new testing infrastructure. We now can simulate web host in Python with a gRPC server to test edge-cases.

4. Validate functions when they are being loaded. Compare parameters in Python with bindings in function.json, error out if they differ.

Addresses issues #11 and #19.

cc @asavaritayal @christopheranderson @elprans


1st1:
Thank you, Asavari!

Issue #4, opened 2018-01-18
HTTP Triggers and Bindings


asavaritayal:
HTTP webhook triggers and HTTP output bindings will use  request and response objects to represent the HTTP messaging. 

#### Request Object
The **request** object has the following attributes.

**Attribute** | **Data Type** | **Description**
-------------|----------|----------------
method | str | HTTP method of the request
url | str | Url of the request
headers | dict | HTTP headers sent with the request
data | dict | Form data parameters
params | dict | Query string parameters
body | bytes | Raw HTTP request body

#### Response Object
The **response** object has the following attributes.

**Attribute** | **Data Type** | **Description**
-------------|----------|----------------
headers | dict | Response headers
status_code | int | HTTP response status code
content | str, bytes, dict | Contents of the response

Similar to any other binding, the HTTP request and response objects can be accessed using the name attribute from function.json.

Example - [https://github.com/Azure/azure-functions-python-worker/tree/master/Samples/HttpTrigger1](https://github.com/Azure/azure-functions-python-worker/tree/master/Samples/HttpTrigger1)

1st1:
I'm wondering if we actually want to pass `Response` object as an argument.  It seems that Azure Java allows users to just return it, which would look like this in Python:

```python
def main(req: Request):
    return Response(status_code=404)
```

There's really no requirement for us to instantiate the `Response` object outside of the `main()` function.  I believe that JavaScript worker allows to simply return `{status: 404}` hash.

brettcannon:
I say just return it. Pre-allocating is not worth it and feels like boilerplate overhead in the function signature to me.

1st1:
Yep, working on a PR to implement that right now.

asavaritayal:
What if the the HttpResponse is not bound to the method return value? Can the user still set its value in the function?

1st1:
Yes, you can declare it as an "out" binding with a different name from $return, and it will work too.

1st1:
Everything here is now implemented in the `dev` branch.

PR #27, opened 2018-02-21
Implement HttpResponse support (see issue #4)

1st1:
See issue #4.

### DIFF
--- a/azure/worker/tests/functions/return_http_404/function.json
+++ b/azure/worker/tests/functions/return_http_404/function.json
@@ -1,0 +1,17 @@
+{
+  "scriptFile": "main.py",
+  "disabled": false,
+  "bindings": [
+    {
+      "authLevel": "anonymous",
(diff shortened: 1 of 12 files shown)


asavaritayal:
Since we're beyond the basic prototype now, let's have @christopheranderson and @pragnagopa review the checkins. Would be good for us to get more familiar with the code and design patterns.

1st1:
Ping.

azure/worker/tests/functions/return_http_404/function.json
```diff
@@ -0,0 +1,17 @@
+{
```
pragnagopa: Nit: Move HttpTrigger tests to folder: HttpTrigger.  Will make it easier to navigate when you add more tests for other triggers

1st1: Good idea! I'll move them in a separate PR.

PR #31, opened 2018-02-22
Fix mock-host test runner to fail & cleanup properly.

1st1:


azure/worker/testutils.py
```diff
@@ -267,7 +267,23 @@ def __init__(self, scripts_dir):
...
+        done, pending = await asyncio.wait(
+            [self._host._connected_fut, self._worker_task],
+            return_when=asyncio.FIRST_COMPLETED)
+
+        try:
+            if self._worker_task in done:
+                self._worker_task.result()
+
+            if self._host._connected_fut not in done:
+                raise RuntimeError('could start a worker thread')
```
elprans: *could not*

PR #48, opened 2018-02-23
Decode incoming gRPC values according to spec in function.json.

1st1:


tests/http_functions/trigger/main.py
```diff
@@ -0,0 +1,3 @@
+def main(req):
+    print("AAAAAAAAAAAAA", flush=True)
```
elprans: Did you intend to commit this in this PR?

1st1: Good catch

PR #54, opened 2018-02-27
Initial support for Blob binding

elprans:
Input bindings are always passed as an `InputStream` object which
implements `BufferedIOBase` and is a read-only non-seekable stream.

Output parameters can be strings, bytes, or file-like objects.

### DIFF
--- a/azure/functions/_abc.py
+++ b/azure/functions/_abc.py
@@ -1,4 +1,5 @@
 import abc
+import io
 import typing
 
 
@@ -87,4 +88,11 @@ class TimerRequest(abc.ABC):
     @property
     @abc.abstractmethod
     def past_due(self) -> bool:
+        pass
+
+
+class InputStream(io.BufferedIOBase, abc.ABC):
+
+    @abc.abstractmethod
+    def read(self, size=-1) -> bytes:
         pass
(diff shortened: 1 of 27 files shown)


azure/worker/type_converters.py
```diff
@@ -107,3 +107,46 @@ def from_proto(cls, data: protos.TypedData) -> typing.Any:
...
+
+class BlobConverter(type_meta.InConverter,
+                    type_meta.OutConverter,
+                    binding=type_meta.Binding.blob):
+
+    @classmethod
+    def check_python_type(cls, pytype: type) -> bool:
+        return (issubclass(pytype, (str, bytes, bytearray,
+                                    azf.InputStream) or
+                hasattr(pytype, 'read')))
```
1st1: Also check if `pytype.read` is a callable? Why is duck typing a good idea here?

azure/functions/_abc.py
```diff
@@ -88,3 +89,14 @@ class TimerRequest(abc.ABC):
...
+
+
+class InputStream(io.BufferedIOBase, abc.ABC):
+
+    @abc.abstractmethod
+    def read(self, size=-1) -> bytes:
+        pass
+
+    @abc.abstractmethod
+    def read1(self, size=-1) -> bytes:
```
1st1: Why do we need `read1`?

azure/worker/type_converters.py
elprans: Duck typing is to support `file-like`.  There's no standard ABC for that, but I can add one if you think that would be cleaner.

azure/functions/_abc.py
elprans: I was under the false impression that `BufferedIOBase` required `read1` to be implemented.  Removed.

Issue #67, opened 2018-03-02
Pass pytype to InConverter.from_proto


christopheranderson:
Problem:

In order to support multiple types for input/trigger bindings in the future, we must know the pytype of the argument for that given binding. 

Proposed solution:

Pass the `pytype` to the `InConverter.from_proto` method

https://github.com/Azure/azure-functions-python-worker/blob/dev/azure/worker/bindings/meta.py#L132

PR #71, opened 2018-03-05
Pass Python type to type converters.

1st1:
This allows to implement custom type encoding/decoding logic
depending on parameter annotatition type.

Closes issue #67.

### DIFF
--- a/azure/worker/functions.py
+++ b/azure/worker/functions.py
@@ -7,6 +7,12 @@ from . import bindings
 from . import protos
 
 
+class ParamTypeInfo(typing.NamedTuple):
+
+    binding_name: str
+    pytype: typing.Optional[type]
+
+
 class FunctionInfo(typing.NamedTuple):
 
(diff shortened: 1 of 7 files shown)


azure/worker/functions.py
```diff
@@ -16,9 +22,9 @@ class FunctionInfo(typing.NamedTuple):
     requires_context: bool
     is_async: bool
 
-    input_binding_types: typing.Mapping[str, str]
-    output_binding_types: typing.Mapping[str, str]
```
christopheranderson: this looks unintentional, we renamed these *_binding_types

christopheranderson: nvm

1st1: Yeah, the new `ParamTypeInfo` struct now contains **both** "binding type name" and "python type", so `input_types` name is a bit more descriptive than `input_binding_types` now.

PR #82, opened 2018-03-15
changed how python worker is started

christopheranderson:
This changes the startup to use a bash script.

This means it won't work on Windows (since most folks don't have cygwin/bash installed). We need to address this eventually. It does work with WSL (which is what I tested on).

### DIFF
--- a/azure/worker/testutils.py
+++ b/azure/worker/testutils.py
@@ -466,7 +466,7 @@ def popen_webhost(*, stdout, stderr, script_root=FUNCS_PATH, port=None):
     extra_env = {
         'AzureWebJobsScriptRoot': str(script_root),
         'workers:config:path': str(worker_path),
-        'workers:python:path': str(worker_path / 'python' / 'worker.py'),
+        'workers:python:path': str(worker_path / 'python' / 'start.sh'),
         'host:logger:consoleLoggingMode': 'always',
     }
 
--- a/python/start.sh
+++ b/python/start.sh
@@ -1,0 +1,25 @@
+#! /bin/bash
+set -x
+
+# Directory name for start.sh
+DIR="$(dirname $0)"
+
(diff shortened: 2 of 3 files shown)


1st1:
@elprans Please take a look

1st1:
We can detect if it's windows and use the script only for nix. This would keep AppVeyor happy.

christopheranderson:
The function's host would have to detect that, so this probably just breaks app veyor for the time being, if we use it.

christopheranderson:
I'd need to add a PR to the azure-functions-host to have "environment" support.

christopheranderson:
I could also skip this for now and just monkey patch the container image, which is what I'm currently doing. We can discuss tomorrow in the call.

christopheranderson:
FYI - I monkey patched the container to unblock us while we discuss the path forward. No need to actually merge this PR till we chat. 

PR: https://github.com/Azure/azure-functions-docker/blob/master/host/dev-nightly/jessie/amd64/python3.6/Dockerfile#L40

1st1:
IMO it would be better if we did this worker script configuration via env.

Here's one way to do it:

```python
def popen_webhost(*, stdout, stderr, script_root=FUNCS_PATH, port=None):
    # ...

    worker_starter = os.environ.get('PYAZURE_WORKER_STARTER')
    if not worker_path:
        worker_starter = str(worker_path / 'python' / 'start.sh')

    # ...

    extra_env = {
        'workers:python:path': worker_starter
    }
```

Then you can just add `start.sh` to your docker image and configure `PYAZURE_WORKER_STARTER` to point to it.  This way the CI and our local setups will stay unaffected.

1st1:
Closing this PR—we'll add an env variable to configure the host/worker if needed.

PR #101, opened 2018-04-13
Import FunctionRpc.proto as a subtree its repo

elprans:
Added subtree from https://github.com/azure/azure-functions-language-worker-protobuf.
Branch: dev.  Commit: 894bebb70bf89f7f4017ec0a5c1d30c1b8f0f776


azure/worker/bindings/http.py
```diff
@@ -97,7 +97,7 @@ def to_proto(cls, obj: typing.Any, *,
                 http=protos.RpcHttp(
                     status_code=str(status),
                     headers=headers,
-                    is_raw=True,
+                    enable_content_negotiation=True,
```
mhoeger: Could you set this to False? Or to let the user choose True if they want to?

On consuming this change, we'll change the host so that is_raw = !enable_content_negotiation. This means that enable_content_negotiation will default to false (same effects as is_raw = true), and we'd like to let users opt-in to the c# content negotiation if they were using it before.

elprans: Which version of the host contains the inverted logic?  We cannot just flip it, because currently `enable_content_negotiation` is the same field as `is_raw`, so flipping it will cause tests to fail.

mhoeger: No current version of the host contains this logic. I was going to consume new language workers and update the host at the same time. Since that requires more coordination, I can also update the .proto file to accept both `is_raw` and `enable_content_negotiation` fields. 

`is_raw` will be deprecated no matter what, so I was hoping to get rid of it in the same step. But I might be getting ahead of myself :)

elprans: The problem is that you've reused the the field in the proto spec, so we have to wait for a host version that supports the new interpretation of the field.

mhoeger: Looks like we're at a bit of a deadlock since the host would have to consume a newer version of the language worker to work too. I'll update the .proto file to have enable_content_negotiation as a new field instead of a replacement!

pragnagopa: synced up with @mhoeger offline. She will be sending a PR to update the Functions Host soon. Please hold off on this PR until then.

pragnagopa: Also, going forward we will be doing only additive changes to the proto file to make sure we do not break language workers

pragnagopa: @elprans - Changes to runtime are checked into dev branch. Let us know if you see any issues with the tests

elprans: @mhoeger @pragnagopa I don't think `Functions.Binaries.2.0.11709-alpha.zip` is working properly with respect to `enable_content_negotiation`.  Setting it to `False` makes this simple HTTP test fail with 406:

https://github.com/Azure/azure-functions-python-worker/blob/e504733bb8318a4813af0c274f5173580a1b5ed8/tests/http_functions/return_http/main.py#L4-L6

pragnagopa:
@mhoeger can you please take a look and confirm the host version that has the changes?

azure/worker/bindings/http.py
mhoeger: @elprans could you try [2.0.11737](https://ci.appveyor.com/project/appsvc/azure-webjobs-sdk-script-y8o14/build/2.0.11737/job/r1d0pxmtkib47bc4/artifacts)? This is the latest version of the host that we will be deploying!

https://github.com/Azure/azure-functions-host/releases

elprans: @mhoeger  2.0.11737 doesn't work either.  `enable_content_negotiation=False` causes HTTP function requests fail with 406.

mhoeger: @elprans i'm looking into it to try to figure out what's going on. It's strange because the change for enable_content_negotiation essentially just reverses the if logic: https://github.com/Azure/azure-functions-host/pull/2696/files

are you seeing `enable_content_negotiation=True` work?

elprans: @mhoeger Yes, `enable_content_negotiation=True` works fine.

elprans: Looking at the host PR, I actually think that the logic has not been reversed in all places, e.g line 166 in `src/WebJobs.Script/Binding/Http/HttpBinding.cs`

mhoeger: Sorry I'm very confused - line 166 is the only place the logic is reversed? (note: the ScriptObjectResult will throw a 406, but RawScriptResult will not)
image

elprans: Sorry, this is a mistake on my end.  Didn't flip `enable_content_negotiation` back to `False` when bumping to 2.0.11737.  I amended the PR.

mhoeger: Sounds great. Thanks! You actually helped us catch an unwanted revert in the master branch of functions-host as well, so thank you :)

pragnagopa:
This needs to be merged as well for the current release. @elprans - anything else pending here?

mhoeger:
@elprans thank you! :)

Issue #110, opened 2018-04-26
Pin version of grpc


pragnagopa:
Is it possible to include a specific version of grpc in Setup script [setup.py#L151](https://github.com/Azure/azure-functions-python-worker/blob/dev/setup.py#L151)? We would like to control when we take dependency on grpc updates

pragnagopa:
Sorry missed concluding on the version. We do want to stick to 1.0.x to be consistent worth other language workers. Tying the host and language workers version numbers might easier to track but does not make sense and they are version independently

elprans:
Fixed in dev.

PR #111, opened 2018-04-26
Pin the dependencies and set the version to 2.0.0 alpha 2 dev

elprans:
Fixes: #110

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,7 @@ class webhost(distutils.cmd.Command):
 
 setup(
     name='azure-functions-python',
-    version='0.0.1',
+    version='2.0.0a2.dev0',
     description='Python Support for Azure Functions',
     classifiers=[
         'License :: OSI Approved :: MIT License',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -148,15 +148,15 @@ def run(self):
     packages=['azure.functions', 'azure.worker', 'azure.worker.protos'],
     provides=['azure.functions'],
     install_requires=[
-        'grpcio',
-        'grpcio-tools',
+        'grpcio~=1.9.1',
```
pragnagopa: can we move to latest version: 1.11.0?

setup.py
```diff
@@ -132,7 +132,7 @@ def run(self):
 
 setup(
     name='azure-functions-python',
-    version='0.0.1',
+    version='2.0.0a2.dev0',
```
pragnagopa: we are still in preview. Lets just update the minor version for this

setup.py
pragnagopa: Can we use =

elprans: `~` there pins the major.minor, but allows updating to the last patch version, which is generally better as that allows picking up non-breaking security fixes. Is absolute version pinning a requirement?

setup.py
elprans: Are you referring to the `dev` suffix?  It's a good idea to have it in between the tags so that we can better identify which version a user reporting an issue is running.

setup.py
pragnagopa: Yea. Makes sense. pinning to a major version should be ok.

setup.py
pragnagopa: I was referring to 2.0

pragnagopa: We will be releasing language -workers and host independently. We will not be able to keep their versions in sync.

elprans: I'm not talking about keeping the *exact* version in sync, but just the major version as an epoch.  python-worker v2 is basically "python support for azure functions v 2.0".

Issue #109, opened 2018-04-26
Increase grpc max message size


pragnagopa:
Functions Host is updated to send command line argument: [--grpcMaxMessageLength ](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Rpc/ProcessManagement/DefaultWorkerProcessFactory.cs#L35)
Python worker needs to be updated to set this value for grpc.max_send_message_length and grpc.max_receive_message_length on the channel

pragnagopa:
@1st1 / @elprans  - We are rolling out functions runtime soon.  It would be good to have the change on the worker sooner to support working with large blobs

elprans:
@pragnagopa Can you point me to the Host artifact that contains this please?

elprans:
Just to confirm, does 11709 also contain the `enable_content_negotiation` (#101) bit?

pragnagopa:
Yes. You can use these for proto changes as well!

PR #112, opened 2018-04-26
Add support for grpcMaxMessageLength command line argument

elprans:
Fixes: #109

### DIFF
--- a/azure/worker/dispatcher.py
+++ b/azure/worker/dispatcher.py
@@ -35,7 +35,7 @@ class Dispatcher(metaclass=DispatcherMeta):
     _GRPC_STOP_RESPONSE = object()
 
     def __init__(self, loop, host, port, worker_id, request_id,
-                 grpc_connect_timeout):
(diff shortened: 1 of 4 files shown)


azure/worker/dispatcher.py
```diff
@@ -319,7 +321,12 @@ def __run_sync_func(self, invocation_id, func, params):
             _invocation_id_local.v = None
 
     def __poll_grpc(self):
-        channel = grpc.insecure_channel(f'{self._host}:{self._port}')
+        options = []
+        if self._grpc_max_msg_len:
+            options.append(('grpc.max_message_length', self._grpc_max_msg_len))
```
pragnagopa: On c#, setting max message length is deprecated. We had to set max receive length and max send length. Same might be true in python. You need to set grpc.max_send_message_length and grpc.max_receive_message_length

elprans: Fixed.

ahmelsayed:
can you get this in soon since dev is currently broken with host dev?

PR #116, opened 2018-05-04
Rename __init__.pyi to __init__.py

dend:


elprans:
We want the top-level `azure` package to be a pure namespace package per PEP 420.  `__init__.pyi` is there for the `mypy` static type checker.

Issue #123, opened 2018-05-14
Consume latest proto file


pragnagopa:
heads up: host will be updated to use latest proto file from https://github.com/Azure/azure-functions-language-worker-protobuf soon.
 Please update python worker to use the latest and let us know if there are any blockers.

pragnagopa:
headers field in RpcHttp is updated to <string, object> pairs instead of <string, string>. This is going to be breaking change.

pragnagopa:
Funtions Runtime artifacts with latest protobuf changes: [2.0.11782-tkfwfyhd artifacts](https://ci.appveyor.com/project/appsvc/azure-webjobs-sdk-script-y8o14/build/2.0.11782-tkfwfyhd/job/7xxo9s73epmn9vx0/artifacts).
Let me know if you cannot access it.

PR #130, opened 2018-05-23
Update protobuf definitions

elprans:
Updated subtree from https://github.com/azure/azure-functions-language-worker-protobuf.
Branch: dev.
Commit: aac95f236681ee11d81d220a88b3cbbd5470380a

Fixes: #123.

### DIFF
--- a/azure/worker/protos/_src/README.md
+++ b/azure/worker/protos/_src/README.md
@@ -17,7 +17,7 @@ From within the Azure Functions language worker repo:
 3.	Add new path in language worker repo to .gitignore file
     -   In .gitignore, add path in language worker repo
 4.	Finalize with commit
-    -	`git commit -m “Added subtree from https://github.com/azure/azure-functions-language-worker-protobuf. Branch: <version branch>. Commit: <latest protobuf commit hash>”`
+    -	`git commit -m "Added subtree from https://github.com/azure/azure-functions-language-worker-protobuf. Branch: <version branch>. Commit: <latest protobuf commit hash>"`
     -	`git push`
 
 ## Pulling Updates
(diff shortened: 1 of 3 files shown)


azure/worker/bindings/http.py
```diff
@@ -142,10 +145,17 @@ def from_proto(cls, data: protos.TypedData, *,
                 f'unsupported HTTP body type from the incoming gRPC data: '
                 f'{body_rpc_type}')
 
+        headers = {}
+        if data.http.headers:
+            for header_name, header_typed_value in data.http.headers.items():
```
pragnagopa: add / update http tests to include request headers

elprans: Headers are covered by the unit tests already.  Do you mean something specific?

pragnagopa: Yes. Test headers with different types for values

elprans: Not sure I follow.  We test functions by making HTTP requests, and per spec the headers are always strings. Can you give an example please?

pragnagopa: Here is an example Keep-Alive: 300

elprans: > Here is an example Keep-Alive: 300 instead of Keep-Alive: "300"

But these are the same thing per HTTP spec.  In fact, `requests` **requires** that the headers dictionary only contains strings.

pragnagopa: Yup. I think you are right. If the requirement for headers to be <string, string>, we should not be changing RpcHttp in proto file. Will get back to you on this.

pragnagopa: I have reverted the changes to proto file. Please update the PR just to pull the latest changes. No other code changes needed. Thanks!

Issue #140, opened 2018-06-04
Rename azure.worker to azure.functions.worker


asavaritayal:
For publishing to PyPI.

elprans:
We cannot move the worker under `azure.functions` as we are separating the packages, and `azure.functions` is not a pure namespace.  I suggest renaming the worker to `azure._functions.worker` instead.

PR #145, opened 2018-06-07
Move azure.worker to azure.functions_worker

elprans:
Fixes: #140


tests/test_blob_functions.py
```diff
@@ -1,6 +1,6 @@
 import time
 
-from azure.worker import testutils
+from azure._functions.worker import testutils
```
pragnagopa: _functions? is this intentional instead of azure.functions.worker?

asavaritayal: From @elprans (https://github.com/Azure/azure-functions-python-worker/issues/140) - 
> We cannot move the worker under azure.functions as we are separating the packages, and azure.functions is not a pure namespace. I suggest renaming the worker to azure._functions.worker instead.

pragnagopa: would azure.functionsWorker work?

pragnagopa: Also, can you please clarify what is pure namespace in python? So we have azure.functions and azure.functions.worker does it mean they are related?

elprans: A pure namespace is a package that only contains other packages (no classes).  Since `azure.functions` contains classes, it's not a pure namespace.

elprans: Changed to `azure.functions_worker`

PR #148, opened 2018-06-18
Split ABC definitions into a standalone repo

elprans:


setup.py
```diff
@@ -164,9 +164,9 @@ def run(self):
 
 
 setup(
-    name='azure-functions-python',
+    name='azure-functions-worker',
```
pragnagopa: rename to match the package azure.functions_worker

setup.py
```diff
@@ -164,9 +164,9 @@ def run(self):
...
     version='1.0.0a2.dev0',
-    description='Python Support for Azure Functions',
+    description='Python Worker for Azure Functions Runtime',
```
pragnagopa: Nit: Python Language Worker for Azure Functions Host

setup.py
```diff
@@ -178,18 +178,17 @@ def run(self):
...
+    packages=['azure.functions_worker',
               'azure.functions_worker.protos',
               'azure.functions_worker.bindings'],
-    provides=['azure.functions'],
     setup_requires=[
         'grpcio~=1.9.1',
         'grpcio-tools~=1.9.1',
     ],
     install_requires=[
         'grpcio~=1.9.1',
```
pragnagopa: update to latest grpc 1.12.0

setup.py
elprans: Python package names customarily have dots and underscores replaced by dashes.  See, for example, https://pypi.org/project/azure-mgmt-storage/

PR #171, opened 2018-08-06
Adding .NET install to yaml

fabiocav:


pragnagopa:
@elprans - 2.1 is the latest dotnet version see here https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/csharp.rb#L418 , which is now 2.1.300

elprans:
@pragnagopa Not 2.1.300-final, though, only the -preview build has been published: see: https://travis-ci.org/Azure/azure-functions-python-worker/builds/412874129#L716

elprans:
> `dotnet --version`
> `2.1.300-rc1-008673`

pragnagopa:
Let's try switching to AppVeyor first. If we still need travis, please create an issue here: https://github.com/travis-ci/travis-build/

fabiocav:
Based on the new logic, 2.1 is supposed to be interpreted as the latest 2.1 version, which technically is supposed to be 2.1.2 runtime/2.1.302 SDK. Based on what I'm seeing though, people have not been successful with a Travis CI setup without an explicit install step, which is what I've pointed to in the other PR (and what we use in our runtime CI builds)

fabiocav:
@elprans still curious, what was the problem (if there was one) with the explicit installation step we mentioned in the other PR?

pragnagopa:
@elprans  - Can you try the install script https://github.com/Azure/azure-functions-host/blob/dev/dotnet-install.sh ?

elprans:
Install script works. I integrated it with other CI fixes in #173. Thanks!

fabiocav:
Good to hear! Thank you!

PR #193, opened 2018-09-11
Update host to 2.0.12080

pragnagopa:


elprans:
Specifically, we need a mechanism to specify the location of the worker config.

pragnagopa:
This is expected as https://github.com/Azure/azure-functions-host/issues/3390 is fixed.
environment variable languageWorkers:python:workerDirectory needs to point to a directory that contains worker.config.json

pragnagopa:
Ping! This runtime is going to production soon.

elprans:
@pragnagopa  Is there a list of breaking changes?  I'm getting:

```Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexingException: Error indexing method 'Functions.queue_trigger' ---> System.InvalidOperationException: Storage account 'AzureWebJobsStorage' is not configured.```

pragnagopa:
There are no intentional breaking changes in this build  https://github.com/Azure/app-service-announcements/issues/136
@elprans - Does the issue repro locally? Should we update the CLI?

pragnagopa:
@brettsam - Any breaking changes around connection strings?

brettsam:
There were some changes to default connection strings, but I don't believe storage was touched. @elprans -- can you share what your function.json looks like? Do you have an AzureWebJobsStorage app setting in your app?

pragnagopa:
I verified AzureWebJobsStorage  is set in AppVeyor

elprans:
I'm testing locally.  All functions which rely on storage are failing with this host version.  Example function.json:

https://github.com/Azure/azure-functions-python-worker/blob/5d62e629ad36bcb5125adface0a57b3e5f50446a/tests/blob_functions/get_blob_bytes/function.json#L1-L23

Connections strings are set as environment variables.

pragnagopa:
Just realized we need to move to latest extensions. Updated the PR

elprans:
That doesn't seem to have helped, I still get the error.

pragnagopa:
Updating extensions is orthogonal to the function indexing issue. @brettsam and I are looking into it to see if it reproes on our end

brettsam:
I just got this set up and tried with and without AzureWebJobsStorage as a connection string -- and it worked. I'm trying to pull this PR now and run the tests.

elprans:
@pragnagopa I think I got it working in #204.

Issue #192, opened 2018-09-08
Writing multiple messages to a single queue


asavaritayal:
I tried the following function - 

```python
import azure.functions as func

def main(req: func.HttpRequest,
         msg: func.Out[func.QueueMessage]) -> str:

    messages = ("apple", "banana", "cherry")

    msg.set(messages)

    return ""
```

```json
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "name": "req",
      "direction": "in",
      "type": "httpTrigger",
      "authLevel": "anonymous"
    },
    {
      "name": "msg",
      "direction": "out",
      "type": "queue",
      "queueName": "outqueue",
      "connection": "AzureWebJobsStorage"
    },
    {
      "name": "$return",
      "direction": "out",
      "type": "http"
    }
  ]
}

```

Received the error below - 
```bash
[9/8/18 12:35:37 AM] Executing 'Functions.Outputs' (Reason='This function was programmatically called via the host APIs.', Id=d487ce2b-7d34-43e7-958c-32fdcbd2110c)
[9/8/18 12:35:37 AM] Executed 'Functions.Outputs' (Failed, Id=d487ce2b-7d34-43e7-958c-32fdcbd2110c)
[9/8/18 12:35:37 AM] System.Private.CoreLib: Exception while executing function: Functions.Outputs. System.Private.CoreLib: Result: Failure
[9/8/18 12:35:37 AM] Exception: TypeError: unable to encode outgoing TypedData: unsupported type "queue" for Python type "tuple"
[9/8/18 12:35:37 AM] Stack:   File "/Users/asavaritayal/python_docs_testing/env1/lib/python3.6/site-packages/azure/functions_worker/dispatcher.py", line 282, in _handle__invocation_request
[9/8/18 12:35:37 AM]     pytype=out_type_info.pytype)
[9/8/18 12:35:37 AM]   File "/Users/asavaritayal/python_docs_testing/env1/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 286, in to_outgoing_proto
[9/8/18 12:35:37 AM]     f'unable to encode outgoing TypedData: '
[9/8/18 12:35:37 AM] .
```

I thought we were supporting multiple outputs via a tuple. Any idea what's wrong? @1st1 @elprans

PR #202, opened 2018-09-20
Allow returning multiple messages in a queue binding

elprans:
Fixes: #192

### DIFF
--- a/azure/functions_worker/bindings/meta.py
+++ b/azure/functions_worker/bindings/meta.py
@@ -1,10 +1,12 @@
 import abc
+import collections.abc
 import datetime
 import enum
 import json
 import typing
 
 from .. import protos
+from .. import typing_inspect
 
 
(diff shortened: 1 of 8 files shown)


azure/functions_worker/bindings/meta.py
```diff
@@ -211,6 +213,28 @@ def to_proto(cls, obj: typing.Any, *,
         pass
 
 
+def is_iterable_type_annotation(annotation: type, pytype: type) -> bool:
+    is_iterable_anno = (
+        typing_inspect.is_generic_type(annotation) and
+        issubclass(typing_inspect.get_origin(annotation),
+                   collections.abc.Iterable)
+    )
```
1st1: Guard against `str` and `bytes`? 😢

1st1: Ah, never mind, the `is_generic_type` call probably is enough.

Issue #221, opened 2018-10-11
Functions hangs when trying to read stream from URL


priyaananthasankar:
This issue happens on my Mac locally. I am unable to isolate the cause.
Attached is the zip file of my function, it inferences Tensorflow InceptionV3 model

Added sample code in __init__.py to read from img url (followed predict shark example) but when running the function, it hangs - breakpoint revealed it hangs while trying to do urllib read:

```
image_url = req.params.get('img')
image_data = urllib.request.urlopen(image_url).read()
```

Unzip the attached file and just create a virtual environment and pip install to see the results.
[TEST.zip](https://github.com/Azure/azure-functions-python-worker/files/2467454/TEST.zip)

This problem was steadily reproducible if a wrong file path was given to tensorflow's File open method,

 outside of functions it fails, but within functions it hangs.

priyaananthasankar:
Identified issue as being 403 forbidden when trying to read an image url without proper headers. I identified this by trying this out with Python shell which clearly tells me 403 forbidden, but python functions just hangs. Here is a snippet of code to try out:
**Scenario 1**

```
user_agent_header = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/535.19 (KHTML, like Gecko) Ubuntu/12.04 Chromium/18.0.1025.168 Chrome/18.0.1025.168 Safari/535.19'
**req_img = urllib.request.Request(image_url, headers={'User-Agent': user_agent_header})**
response_img = urllib.request.urlopen(req_img)
image_data = response_img.read()
```
Now remove the user_agent header in the line in bold above and function hangs.

**Scenario 2**

Try to read a file with wrong path using Tensorflow:

For eg: model_file is the absolute path to a model, just make it an invalid path. Function hangs.
```
with tf.gfile.FastGFile(model_file, 'rb') as f:
    graph_def = tf.GraphDef()
    graph_def.ParseFromString(f.read())
    _ = tf.import_graph_def(graph_def, name='')
```

elprans:
@priyaananthasankar Can you elaborate on what you mean by "function hangs".  The scenarios you listed generate a Python exception, which is propagated to the host as a `Microsoft.Azure.WebJobs.Host.FunctionInvocationException`, and the function execution is terminated.  We have a test for that in the suite:

[...]

PR #225, opened 2018-10-15
Prevent user exception serialization from failing

elprans:
Currently, if an unhandled exception in user function cannot be
serialized properly, the whole worker crashes.  Prevent that from ever
happening.

Fixes: #221

### DIFF
--- a/tests/http_functions/unhandled_unserializable_error/main.py
+++ b/tests/http_functions/unhandled_unserializable_error/main.py
@@ -1,0 +1,10 @@
+import azure.functions as func
+
+
+class UnserializableException(Exception):
+    def __str__(self):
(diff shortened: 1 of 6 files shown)


tests/http_functions/unhandled_unserializable_error/main.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class UnserializableException(Exception):
+    def __str__(self):
+        raise RuntimeError('cannot serialize me')
+
+
+def main(req: func.HttpRequest) -> str:
+    raise UnserializableException('foo')
+    return 'OK'
```
1st1: drop return?

Issue #246, opened 2018-11-29
Python worker 1.0.0b1 is a breaking change from 1.0.0a6


pragnagopa:
Worker.py needs to be backward compatible with  1.0.0a6

asavaritayal:
1.0.0b1 has been held off from the release due to the possibility of a breaking change.

PR #247, opened 2018-11-29
Restore compatibility with pre-b1 releases of worker in worker.py

elprans:
Fixes: #246

### DIFF
--- a/python/worker.py
+++ b/python/worker.py
@@ -1,4 +1,46 @@
-from azure.functions_worker import main
+"""Main entrypoint."""
+
+
+try:
+    from azure.functions_worker.main import main
+
+except ImportError:
+    # Compatibility with hard-bundled pre-beta worker versions in
+    # deployed function apps.
+    import argparse
+    import traceback
+
+    def parse_args():
+        parser = argparse.ArgumentParser(
+            description='Python Azure Functions Worker')
+        parser.add_argument('--host')
+        parser.add_argument('--port', type=int)
+        parser.add_argument('--workerId', dest='worker_id')
+        parser.add_argument('--requestId', dest='request_id')
+        parser.add_argument('--log-level', type=str, default='INFO',
+                            choices=['TRACE', 'INFO', 'WARNING', 'ERROR'],)
+        parser.add_argument('--log-to', type=str, default=None,
+                            help='log destination: stdout, stderr, '
+                                 'syslog, or a file path')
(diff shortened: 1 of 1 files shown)


maiqbal11:
We can hold off on this fix for now. @ahmelsayed is going to try to change the `PYTHONPATH` so that we reference the version of the worker we install on the image rather than using the version that the user provides.

ankitkumarr:
@maiqbal11, I think we still may need this change to ensure compatibility with older versions of the CLI.

maiqbal11:
> @maiqbal11, I think we still may need this change to ensure compatibility with older versions of the CLI.

Older versions of the CLI should be using the previously released versions of the `worker.py` and `worker.config.json`. The `PythonWorkerUrl` and `PythonWorkerConfigUrl` variables in the settings pick these up based on the release. For example, the previous `worker.py` in `1.0.0a6` (https://raw.githubusercontent.com/Azure/azure-functions-python-worker/1.0.0a6/python/worker.py) is still available for previous CLI versions. I believe this should ensure compatibility in older CLI versions. Was this the concern that you had or is it something different?

ankitkumarr:
@maiqbal11, 
My concern is about older apps trying to publish with `--build-native-deps`. Say, some app with dependency on a6 tries to publish, we will pull the newest Docker image (with b1) and package their app inside it. While packaging, we may be trying to bundle the new `worker.py` with their venv's old versions of `azure-functions-worker` (a6). Given that new `worker.py` is incompatible with the old `azure-functions-worker`, they may face issues.
Does that make sense?

maiqbal11:
> @maiqbal11,
> My concern is about older apps trying to publish with `--build-native-deps`. Say, some app with dependency on a6 tries to publish, we will pull the newest Docker image (with b1) and package their app inside it. While packaging, we may be trying to bundle the new `worker.py` with their venv's old versions of `azure-functions-worker` (a6). Given that new `worker.py` is incompatible with the old `azure-functions-worker`, they may face issues.
> Does that make sense?

@asavaritayal , this is a valid concern about breaking existing customers since we'll be using the worker installed in the base image going forward which could be incompatible with the one the from the previous releases. Can we post an announcement regarding this?

elprans:
@maiqbal11 AFAIK we have not made any breaking changes for a while, so existing functions should work with the new worker versions without an issue.

maiqbal11:
Merging this is in as we'll need it to do a new release with the fixes that were in `1.0.0b1` but were rolled back.

PR #283, opened 2019-01-15
Adding Azure DevOps CI

maiqbal11:
Adding code for Azure DevOps

### DIFF
--- a/.ci/linux_devops_dotnet.sh
+++ b/.ci/linux_devops_dotnet.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-
-set -e -x
-mkdir .dotnet
-wget https://raw.githubusercontent.com/dotnet/cli/80d542b8f4eff847a0f72dc8f2c2a29851272778/scripts/obtain/dotnet-install.sh
-chmod +x dotnet-install.sh
-./dotnet-install.sh --version ${DOTNET_VERSION} --install-dir .dotnet
-.dotnet/dotnet --info
--- a/.ci/linux_devops_dotnet.sh
+++ b/.ci/linux_devops_dotnet.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -e -x
+mkdir .dotnet
+wget https://raw.githubusercontent.com/dotnet/cli/80d542b8f4eff847a0f72dc8f2c2a29851272778/scripts/obtain/dotnet-install.sh
+chmod +x dotnet-install.sh
+./dotnet-install.sh --version ${DOTNET_VERSION} --install-dir .dotnet
+.dotnet/dotnet --info
--- a/.ci/win_devops_tests.bat
+++ b/.ci/win_devops_tests.bat
@@ -1,0 +1,3 @@
+python .ci\devops_setup_worker.py > worker_path
+set /p PYAZURE_WORKER_DIR=<worker_path
+python setup.py test
--- a/.ci/win_devops_tools.bat
+++ b/.ci/win_devops_tools.bat
@@ -1,3 +1,3 @@
 choco install azure-functions-core-tools --pre
-dotnet --info
+dotnet --version
 python --version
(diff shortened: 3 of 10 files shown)


.ci/linux_devops_dotnet.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
```
pragnagopa: Is installing dotnet explicitly needed?

.ci/win_devops_tests.bat
```diff
@@ -0,0 +1,3 @@
+python .ci\devops_setup_worker.py > worker_path
+set /p PYAZURE_WORKER_DIR=<worker_path
+python setup.py test
```
pragnagopa: Missed defining environment variables? Similar to linux_devops_tests.sh?

.ci/win_devops_tools.bat
```diff
@@ -0,0 +1,3 @@
+choco install azure-functions-core-tools --pre
+dotnet --info
```
pragnagopa: dotnet --info is too verbose. Use dotnet --version instead

azure-pipelines.yml
```diff
@@ -0,0 +1,84 @@
+trigger:
+- dev
```
pragnagopa: what about master branch?

azure-pipelines.yml
```diff
@@ -0,0 +1,84 @@
...
+    displayName: 'Build'
+  - task: BatchScript@1
+    inputs:
+      filename: .ci\win_devops_tests.bat
+    env:
+      AzureWebJobsStorage: $(WindowsStorageConnectionString)
+      AzureWebJobsCosmosDBConnectionString: $(WindowsCosmosDBConnectionString)
+      AzureWebJobsEventHubConnectionString: $(WindowsEventHubConnectionString)
+      AzureWebJobsServiceBusConnectionString: $(WindowsServiceBusConnectionString)
+    displayName: 'Test'
```
pragnagopa: Nit: Test_Windows?

azure-pipelines.yml
```diff
@@ -0,0 +1,84 @@
...
+    displayName: 'Build'
+  - bash: |
+        chmod +x .ci/linux_devops_tests.sh
+        .ci/linux_devops_tests.sh
+    env:
+      LINUXSTORAGECONNECTIONSTRING: $(LinuxStorageConnectionString)
+      LINUXCOSMOSDBCONNECTIONSTRING: $(LinuxCosmosDBConnectionString)
+      LINUXEVENTHUBCONNECTIONSTRING: $(LinuxEventHubConnectionString)
+      LINUXSERVICEBUSCONNECTIONSTRING: $(LinuxServiceBusConnectionString)
+    displayName: 'Test'
```
pragnagopa: Nit: Test_Linux

.ci/win_devops_tests.bat
maiqbal11: This is fine. We don't need to export environment variables for Windows. Linux requires exporting environment variables so child processes can use them.

azure-pipelines.yml
maiqbal11: Didn't add that yet since I wanted to test it out on the dev branch before pushing it out to the master. Now that you mention it: Would be better to add it in this PR before merging into master so that changes goes into master. I'll add it in.

azure-pipelines.yml
maiqbal11: This should be fine. It's easy to see from the context in the yaml file as well as the UI which system is being tested.

.ci/linux_devops_dotnet.sh
maiqbal11: I've changed the yaml file so that it uses an explicit task to install the required dotnet version. We do need to specify the dotnet version we are using as there are multiple version available and we might not want to test against the default.

PR #284, opened 2019-01-15
Removing Travis and AppVeyor CI

maiqbal11:
Azure DevOps CI code is already checked into the dev branch. We should remove Travis and AppVeyor so that we are not maintaining 3 different CI systems simultaneously. Plans are to make all future CI investments using Azure DevOps.


maiqbal11:
> @maiqbal11 Please configure the repo to block on CI status.

Enabled blocking on CI status for `dev` and `master` branches.

maiqbal11:
> Also, it would be nice to continue to have a CI badge if possible.

Planning on adding that for Azure Pipelines and also enabling CI on PRs. Will make those changes before getting this PR in.

maiqbal11:
Closing in favor of #285 so that we can get a full CI build. CI builds from forks are not yet enabled.

PR #300, opened 2019-01-25
Update proto

maiqbal11:
Updating to latest protobuf


maiqbal11:
@elprans, `data_type` will be exposed as part of the `FunctionLoadRequest` when sending `BindingInfo`. You should be able to review quickly to make sure this meets the requirement.

Issue #269, opened 2018-12-27
Python DateTime Error For Blob Events In Event Grid Azure Function


eragonriddle:
# Issue Description
The Azure Event Grid Trigger function in python throws an error on receiving a blob event. The error is related to DateTime format. The date returned by the blob event to the python function cannot be parsed by python as the millisecond value has seven digits. E.g. the date '2018-12-09T23:47:03.285137Z'  cannot be parsed by python as the millisecond value has seven digits which is not recognized in python.

```
Result: Failure
Exception: ValueError: time data '2018-12-12T03:16:34.2191989Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'
Stack:   File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/dispatcher.py", line 250, in _handle__invocation_request
    pytype=pb_type_info.pytype)
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 285, in from_incoming_proto
    trigger_metadata=trigger_metadata)
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/eventgrid.py", line 94, in from_proto
    event_time=cls._parse_datetime(body.get('eventTime')),
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 181, in _parse_datetime
    raise last_error
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 176, in _parse_datetime
    dt = datetime.datetime.strptime(datetime_str, fmt)
  File "/root/.pyenv/versions/3.6.6/lib/python3.6/_strptime.py", line 565, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/root/.pyenv/versions/3.6.6/lib/python3.6/_strptime.py", line 362, in _strptime
    (data_string, format))
```

# Steps to reproduce issue

```bash
#### STEP 1: CREATE AZURE FUNCTION

groupName="<resource_group_name>"
location="westeurope"
functionStorageName="<function_storage_account_name>"
appinsightsKey="<application_insights_instrumentation_key>"
appName="<function_app_name>"

#### CREATE RESOURCE GROUP

az group create --name $groupName --location $location

[...]

PR #306, opened 2019-01-30
Fix parsing of datetime metadata with 7+ digits in second fraction

elprans:
It seems that Azure returns datetime with nanosecond precision, wheras
Python's datetime format only expects microseconds, so strip the extra
digits.

Fixes: #269


maiqbal11:
Would be better if we could use the standard library to support this, but it looks like they have an issue open for this: https://bugs.python.org/issue15443. Looks good otherwise!

asavaritayal:
@elprans is this good to merge?

PR #245, opened 2018-11-29
[do not merge]Include worker.config.json in distribution metadata

elprans:
`worker.config.json` is now included in the worker package distribution
metadata, and can be accessed like this:

    import pkg_resources
    dist = pkg_resources.get_distribution('azure-functions-worker')
    worker_config = dist.get_metadata('worker.config.json')

Fixes: #243
Requires: https://github.com/Azure/azure-functions-host/pull/3369

PR #311, opened 2019-02-01
Create nuget package for worker

maiqbal11:
Creating a nuget package for Python Worker that can be used by core tools and for testing latest code.

### DIFF
--- a/python/run_worker.py
+++ b/python/run_worker.py
@@ -1,0 +1,8 @@
+import platform
+import subprocess
+
+if __name__ == '__main__':
+
+    subprocess.run(['python', '-m', 'pip', 'install', '.'])
+    subprocess.run(['python', 'setup.py', 'build'])
+    subprocess.run(['python', '-m', 'azure.functions_worker'])
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,8 @@ ENV/
 # mypy
 .mypy_cache/
 
+# nuget
+*.nupkg
 
 .testconfig
 .pytest_cache
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -84,7 +84,17 @@ jobs:
       LINUXEVENTHUBCONNECTIONSTRING: $(LinuxEventHubConnectionString)
       LINUXSERVICEBUSCONNECTIONSTRING: $(LinuxServiceBusConnectionString)
     displayName: 'Test'
+  
+- job: Nuget
+  pool:
+      vmImage: 'vs2017-win2016'
+  steps:
+  - task: NuGetCommand@2
+    inputs:
+      command: pack
+      packagesToPack: 'pack\worker.nuspec'
 
-
-
-
+  - task: PublishBuildArtifacts@1
+    inputs:
+      pathtoPublish: '$(Build.ArtifactStagingDirectory)' 
+      artifactName: 'Microsoft.Azure.Functions.PythonWorker'
--- a/pack/worker.nuspec
+++ b/pack/worker.nuspec
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+  <metadata>
+    <id>Microsoft.Azure.Functions.PythonWorker</id>
+    <version>1.0.0-beta3</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Microsoft Azure Functions Python Worker</description>
+    <copyright>© .NET Foundation. All rights reserved.</copyright>
+    <contentFiles>
+      <files include="..\azure\**" buildAction="None" copyToOutput="true" flatten="false"/>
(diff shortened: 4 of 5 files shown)


maiqbal11:
Need to add code to the pipeline for creating nuget package. Will create another PR once that it is done.

python/run_worker.py
```diff
@@ -0,0 +1,8 @@
+import platform
+import subprocess
+
+if __name__ == '__main__':
+
+    subprocess.run(['python', '-m', 'pip', 'install', '.'])
+    subprocess.run(['python', 'setup.py', 'build'])
```
ahmelsayed: does this mean that it'll run pip install, and build on every activation of the python worker? Won't that make it too slow?

elprans: Yes, this doesn't seem like a good idea.  @maiqbal11 why do you need this?

maiqbal11: Following the logic of nuget packaging for other workers, the nuget should be as self-contained as possible. Given that, I didn't want to assume that the packages needed to run the worker were already present in the environment so I'm using pip install to gather them. The alternative would have been to package some kind of virtual environment in the nuget which would end up being platform specific. I don't feel too strongly about this though and am open to ideas.

elprans: @maiqbal11 There has to be another mechanism, like a post-install hook or something.  Doing this in `__main__` adds a lot to the startup time.

maiqbal11: Agree with that. I'll look into alternatives.

maiqbal11: I've found several issues to indicate that NuGet is going to pose some limitations for the intended purposes of this PR (which is to make a self-contained runnable worker distribution):

1. Post-install hooks (in the form of powershell scripts) have been deprecated in NuGet for some time now: https://github.com/NuGet/Home/issues/4318#issuecomment-343255043.
2. NuGet generally is moving towards a purely library packaging format without custom installation logic:  https://github.com/NuGet/Home/issues/5963. 
3. The only existing custom installation logic does not run from command line (only VS): https://github.com/NuGet/Home/issues/4318. 

As an alternative, I also investigated Chocolatey, which is the Windows-standard way of packing applications. Arguably, we are dealing more with an application here more than pure code. This works well for Windows via `choco install` which essentially runs custom installation logic on files packaged in NuGet format. However, Chocolatey is not suppored out-of-the-box on Linux and, even then, support is limited. 

As things stand, we can do one of two things: 

1. Package scripts, separately for Linux and Windows, that would need to be run to set up the environment before running the worker.
2. Use our exisiting PyPi worker package which also installs the required dependencies upon installation. There is a PR out to package the `worker.config.json` with the PyPi package as a resource: https://github.com/Azure/azure-functions-python-worker/pull/245. 

I'd be in favor of the second option since it utilizes the Pythonic mechanism that we already have in place and also neatly takes care of dependencies. With option 1, we'd have some consistency with other workers' packaging but would need to do some platform specific work. 

Would like some more clarity on the intended use cases so we can use the most appropriate mechanism. Thoughts? @ankitkumarr, @elprans, @ahmelsayed, @asavaritayal.

pragnagopa: Thanks for details @maiqbal11 ! yes, I am in favor of merging PR #245 as long we figure out a plan for release either
- non breaking way or
- if it is a breaking change, provide guidance to customers on how to fix themselves.

maiqbal11: @pragnagopa, I believe we should be able to do this in a non-breaking way since @ankitkumarr added the ability to pin to specific docker images: https://github.com/Azure/azure-functions-core-tools/commit/8efec779f5bbe99940433546be885b7497e251c0.

maiqbal11: @ankitkumarr, @ahmelsayed, is using PyPi a feasible solution? If so, I can close this PR + issue and merge #245.

ankitkumarr: @maiqbal11, I think the goal here is to package the Python worker with func CLI. If it's a PyPi platform dependent package, we would have to `pip install` as a post installation task of the func CLI. In that case, we would be taking a dependency on `pip` even for non-Python function users.
I am not sure if that's reasonable. 

Makes me wonder if we should have `func install worker python` kind of model where language workers are extensions of the CLI.

maiqbal11: @ankitkumarr, the `pip install` command can happen when the user selects `python` as the language they want to use. The only dependency then would be to have Python installed on the system - which is a minimum requirement either way.

ankitkumarr: @maiqbal11, in that case, if a user clones a function app, and does `func host start`, it might fail because we were thinking of not including `worker` in `requirements.txt`.

Also, I think we were discussing to not install worker on the user's path so that if user runs `pip freeze`, we do not add that in `requirements.txt`. That way, we don't filter things out during publish. But, if there's no easy way to do this, it might be ok.

maiqbal11: @asavaritayal had suggested that we can attempt to do the `pip install` every time we call `func host start` if the app is Python. This could still be a viable solution given that `pip` will not re-install if it already finds the requirements there. 

I think it's fine that the user's `requirements.txt` includes the worker requirements. We just need to ensure that our versions of these are being used when we deploy to the Docker environment/container.

maiqbal11:
Closing favor of https://github.com/Azure/azure-functions-python-worker/pull/329 so we can get a CI build with proper artifacts.

PR #316, opened 2019-02-07
Always use raw HTTP body

elprans:
`RpcHttp.body` may be reformatted by the Functions Host, and as such,
may not be identical to the original body sent by the client.  This has
numerous problems (e.g invalid Content-Length etc.)  Sidestep the issue
by always using `RpcHttp.rawBody`.

See also: Azure/azure-functions-host#3875
Fixes: #260


maiqbal11:
Ping. @elprans, did you get a chance to see if we can add the HMAC check for the message body?

elprans:
@maiqbal11 I added a test.

Issue #296, opened 2019-01-24
Using multiple outputs as parameters not working


FlippieCoetser:
### Send data to a single output binding
Two options are available to send data to an output binding (HTTP Request / Response example below).

#### Using `func.Out[T]`
a HTTP Response message is produced by using `.set(T)` method on `func.out[T]`.
The parameter `reponse` in `main.py` and name `response` in `function.json` is the same 

#### `main.py`

```python
import logging
import azure

def main(request: azure.functions.HttpRequest,
         response: azure.functions.Out[azure.functions.HttpResponse]):
    HttpResponse = azure.functions.HttpResponse("HTTP Response", status_code=200)
    # Response is provided using .set method on func.Out[T] 
    response.set(HttpResponse)
    return 
```

#### `function.json`

```json
{
  "scriptFile": "main.py",
  "entryPoint": "main",
  "bindings": [
    {
      "name": "request",
      "type": "httpTrigger",
      "direction": "in",
      "methods": [
        "get",
        "post"
      ],
      "authLevel": "function"
    },
    {
      "name": "response",
      "type": "http",
      "direction": "out"
    }
  ]
}
```

#### Using function `return T`
a HTTP Response message is produced by using the `main` functions return value.
Notice the return value `HttpResponse` in `main.py` and name `$return` in `function.json` 

#### `main.py`

```python
import logging
import azure

def main(request: azure.functions.HttpRequest) -> azure.functions.HttpResponse: 
    HttpResponse = azure.functions.HttpResponse("HTTP Response", status_code=200)
    # Response is provided using methods return value 
    return HttpResponse
```

#### `function.json`

```json
{
  "scriptFile": "main.py",
  "entryPoint": "main",
  "bindings": [
    {
      "name": "request",
      "type": "httpTrigger",
      "direction": "in",
      "methods": [
        "get",
        "post"
      ],
      "authLevel": "function"
    },
    {
      "name": "$return",
      "type": "http",
      "direction": "out"
    }
  ]
}
```

#### Using a mix: Function return T and `func.Out[T]`
1. a HTTP Response message is produced by using the `main` functions return value.
2. a message to queue storage is produced by using `.set(T)` method on `func.out[T]`.
3. The parameter `queue` in `main.py` and name `queue` in `function.json` is the same
4. Notice The return value `HttpResponse` in `main.py` and name `$return` in `function.json`

#### `main.py`
```python
import logging
import azure

def main(request: azure.functions.HttpRequest,
[...]

PR #317, opened 2019-02-07
Fix handling of multiple Out params

elprans:
Fixes: #296

### DIFF
--- a/azure/functions_worker/dispatcher.py
+++ b/azure/functions_worker/dispatcher.py
@@ -295,7 +295,7 @@ class Dispatcher(metaclass=DispatcherMeta):
             output_data = []
             if fi.output_types:
                 for out_name, out_type_info in fi.output_types.items():
-                    val = args[name].get()
+                    val = args[out_name].get()
                     if val is None:
                         # TODO: is the "Out" parameter optional?
                         # Can "None" be marshaled into protos.TypedData?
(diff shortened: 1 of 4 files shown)


maiqbal11:
@elprans, is this good to check in?

elprans:
@maiqbal11 Seems like CI has failed with a host error:

```
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.get_queue_blob_message_return ---> System.AggregateException: One or more errors occurred. (The condition specified using HTTP conditional header(s) is not met.) ---> Microsoft.WindowsAzure.Storage.StorageException: The condition specified using HTTP conditional header(s) is not met.
```

I cannot reproduce locally, this seems to be intermittent.  Is there a way to restart the build?

maiqbal11:
> @maiqbal11 Seems like CI has failed with a host error:
> 
> ```
> Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.get_queue_blob_message_return ---> System.AggregateException: One or more errors occurred. (The condition specified using HTTP conditional header(s) is not met.) ---> Microsoft.WindowsAzure.Storage.StorageException: The condition specified using HTTP conditional header(s) is not met.
> ```
> I cannot reproduce locally, this seems to be intermittent. Is there a way to restart the build?

I've re-triggered it. You should be able to navigate to the CI page and re-trigger by choosing "Rebuild" in the `...` dropdown menu at the top right. Does that option show up for you?

PR #325, opened 2019-02-13
Add Function App directory as a top-level module

maiqbal11:
Given the issues we've been getting for absolute imports over time, wanted to propose a change to the way we are currently handling namespaces for our Function App. Basically, the change involves removing `__azure__` as a top-level module and using the Function Application's directory name as a top-level module. Taking for example the following directory structure:

```
FunctionApp
  Function1
    __init__.py
    func_module.py
  HelperCode
    __init__.py
    helper_module.py
```

FunctionApp would be added as a top-level module instead of `__azure__`. We can then publish guidance to allow users to access absolute imports via the following syntaxes (assuming we are importing from `Function1.__init__.py`):

1. `from FunctionApp import HelperCode`
2. `from FunctionApp.HelperCode import helper_module`

Relative import behavior would remain unchanged. This is likely not the equivalent of what users expect from non-Functions Python but it gets us something in the way of supporting absolute imports (without having to be too heavy-handed with import logic).

### DIFF
--- a/azure/functions_worker/dispatcher.py
+++ b/azure/functions_worker/dispatcher.py
@@ -82,8 +82,6 @@ class Dispatcher(metaclass=DispatcherMeta):
 
         self._old_task_factory = self._loop.get_task_factory()
 
-        loader.install()
-
         DispatcherMeta.__current_dispatcher__ = self
         try:
             forever = self._loop.create_future()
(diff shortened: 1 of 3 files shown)


maiqbal11:
\cc @asavaritayal

asavaritayal:
As discussed, `FunctionApp` does not remain consistent across local development and hosting in Azure. We should just call it `__app__`.

Issue #274, opened 2019-01-07
Integration testing for language worker, core tools and docker image


asavaritayal:
Details below

maiqbal11:
We will need to create a setup which can replicate end to end user scenarios: Using core tools to create functions and then publishing them to Linux. We would use the latest core tools, language worker and docker image for this. Will provide further updates once I have further clarity on what all we need to include.

maiqbal11:
Created an issue from core-tools perspective so we can ensure that deployment options do not break when changes are made to worker: https://github.com/Azure/azure-functions-core-tools/issues/971. The eventual goal should be to decouple them entirely and simply expose a config that the core tools use as an entry point.

maiqbal11:
Testing from Core-tools perspective: https://github.com/Azure/azure-functions-core-tools/issues/971

maiqbal11:
OGF Testing for user experience: https://github.com/Azure/azure-functions-core-tools/issues/1040.

maiqbal11:
Should be in a good state with our testing for Python: CI on Python worker, OGF testing and integration testing with Core Tools/Docker. Any improvements at this point should be incremental and can be tracked in new work items.

PR #332, opened 2019-02-20
Add e2e tests - inital

ankitkumarr:
Closes https://github.com/Azure/azure-functions-core-tools/issues/971 and maybe(?) https://github.com/Azure/azure-functions-python-worker/issues/274
[We discussed that Python tests should belong in this repo]

Adding a bunch of e2e tests:

These can be run in two ways-
1. Using Azure pipelines from the `azure-pipelines-nightly.yml` which will be a nightly build.
2. Using a local docker container.

**Using a local docker container**
You would need to fill out the `.ci/e2e/publish_tests/publish_config.json` appropriately with your own azure resources and run these two commands-
```
docker build -f .ci\e2e\running-environments\Docker\test_runner.Dockerfile -t my-test-image <path-to-python-worker>
docker run -v /var/run/docker.sock:/var/run/docker.sock my-test-image
```

I will make a wiki page on the structure, details on the implementation and how to run these once/if this PR is approved.

once ran, and if things go well, you should get an output like this-

capture

\cc @pragnagopa @ahmelsayed @maiqbal11 @fabiocav @elprans @asavaritayal

### DIFF
--- a/.ci/e2e/publish_tests/publish_config.json
+++ b/.ci/e2e/publish_tests/publish_config.json
@@ -1,0 +1,68 @@
+{
+  "azure_service_principal": {
+    "user_id": "${AZURE_SERVICE_PRINCIPAL_USER}",
+    "password": "${AZURE_SERVICE_PRINCIPAL_PASSWORD}",
+    "tenant": "${AZURE_SERVICE_PRINCIPAL_TENANT}"
+  },
+  "docker_setup": {
+    "acr_name": "${ACR_NAME}",
+    "working_dir": "/docker-dir",
+    "dev_image_name": "${ACR_IMAGE_NAME}",
+    "prod_image": "mcr.microsoft.com/azure-functions/python:2.0"
+  },
+  "func_setup": {
+    "func_dev_dir": "/func-dev-dir",
+    "func_dev_url": "https://functionsclibuilds.blob.core.windows.net/builds/2/latest/Azure.Functions.Cli.linux-x64.zip"
+  },
+  "tests": {
+    "working_dir": "/tests-dir",
+    "logs": "/tests-logs",
+    "timeout": "3000s"
+  },
+  "publish_function_app": {
+    "prod_func_prod_docker": {
+      "new_function": {
+        "build_native_deps": "${FUNCTION_APP}",
+        "no_bundler": "${FUNCTION_APP}",
+        "packapp": "${FUNCTION_APP}"
+      },
+      "customer_churn": {
+        "build_native_deps": "${FUNCTION_APP}",
+        "no_bundler": "${FUNCTION_APP}"
+      }
+    },
+    "prod_func_dev_docker": {
+      "new_function": {
+        "build_native_deps": "${FUNCTION_APP}",
+        "no_bundler": "${FUNCTION_APP}",
+        "packapp": "${FUNCTION_APP}"
+      },
+      "customer_churn": {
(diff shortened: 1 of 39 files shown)


ahmelsayed:
wow, this is pretty cool. Great job! I think a wiki page as you mentioned would be great. Maybe you can just go through the process of adding a test case would be very helpful. 
How long does the CI take to run through all the scenarios? also is the parallel script actually able to run? seems like a lot for one machine :)

The only comment I have is that I think you could probably add a `set -e` to non-test scripts to make them exit as soon as they hit an error. It makes things more predictable.

ankitkumarr:
@ahmelsayed, thanks for the feedback!
That sounds great! I will definitely include that in the wiki page.

I didn't record the time to run all in the CI yet, but I think it should take about an hour or two. I was facing some issues due to that bug you [fixed](https://github.com/Azure/azure-functions-core-tools/pull/1022) recently, core-tools deadlocks in the CI and I have pretty long timeouts right now, while Azure Devops times out before my timeouts. But that will be fixed in the next release of core-tools. I did run them all locally, but didn't record the times. I will make an update to increase the timeout in the CI.

Yes, you are right about the parallel script. I included it initially because I figured if someone wanted to run say 4 test cases in parallel, they can comment others out. That's why I made the config file so that they can use different function apps for different tests in order to avoid false positives.  But it might be unnecessary.

Good point about the `set -e`. I think specially with the dev docker setup part, because one would never know otherwise if the Docker image even got updated. I will update the PR to take care of that. Thanks for the suggestion!

.ci/e2e/publish_tests/publish_config.json
```diff
@@ -0,0 +1,68 @@
+{
+  "azure_service_principal": {
+    "user_id": "${AZURE_SERVICE_PRINCIPAL_USER}",
```
pragnagopa: Add helper script to generate spn : https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest

.ci/e2e/publish_tests/publish_config.json
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "docker_setup": {
+    "acr_name": "${ACR_NAME}",
+    "working_dir": "/docker-dir",
+    "dev_image_name": "${ACR_IMAGE_NAME}",
+    "prod_image": "mcr.microsoft.com/azure-functions/python:2.0"
+  },
+  "func_setup": {
+    "func_dev_dir": "/func-dev-dir",
+    "func_dev_url": "https://functionsclibuilds.blob.core.windows.net/builds/2/latest/Azure.Functions.Cli.linux-x64.zip"
```
pragnagopa: Move the url to environment var

ankitkumarr:
all_green

After the new release of the CLI, I am getting all green tests. I think this is ready to be merged. Any additional improvements can be addressed in separate issues/PRs. 

Once merged, I can start the nightly build on the appropriate branch.

@pragnagopa / @maiqbal11 would you mind merging this, I do not have access.

ankitkumarr:
Thanks Pragna for adding me. Looks like the PR CI is hung up. I will close the PR and re-open to kick off another CI.

ankitkumarr:
Here's the documentation: https://github.com/Azure/azure-functions-python-worker/wiki/End-to-End-publishing-tests

PR #353, opened 2019-03-01
Update CI tests for more control

ankitkumarr:
Three changes:

1. Add ability to override the func artifacts link in the CI
2. Add a 5 second delay between publish and trigger test. Because sometimes, the trigger test is too quick and it returns a 404.
3. Change line endings when copying scripts to a docker container.

### DIFF
--- a/.ci/e2e/publish_tests/func_tests_core/customer_churn_app/no_bundler_test.sh
+++ b/.ci/e2e/publish_tests/func_tests_core/customer_churn_app/no_bundler_test.sh
@@ -11,6 +11,7 @@ FUNCTIONS_PYTHON_DOCKER_IMAGE=$4 "$3" azure functionapp publish "$2" --build-nat
 
 if [[ $? -eq 0 ]]
 then
+    sleep 5s
     # https://stackoverflow.com/questions/2220301/how-to-evaluate-http-response-codes-from-bash-shell-script
     STATUS_CODE=$(curl --write-out %{http_code} --silent --output /dev/null https://$2.azurewebsites.net/api/predict)
     verify_status_code $1.result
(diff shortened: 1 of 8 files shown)


.ci/e2e/publish_tests/func_tests_core/customer_churn_app/no_bundler_test.sh
```diff
@@ -11,6 +11,7 @@ FUNCTIONS_PYTHON_DOCKER_IMAGE=$4 "$3" azure functionapp publish "$2" --build-nat
 
 if [[ $? -eq 0 ]]
 then
+    sleep 5s
```
pragnagopa: Instead of 5s delay in the test, add delay 30sec right after func host start. add logging waiting for host to be ready

ankitkumarr: @pragnagopa, these tests do not run `func host start`. They are testing publish actions.

ankitkumarr: The delay is between publish and trying to check the trigger endpoint. It's not necessary, but in some cases I saw a second or two delay

pragnagopa: Sorry. I meant right after publish. Give 10sec-15sec before pinging

pragnagopa: never mind. i remember the flow now. each test is publish and ping?

ankitkumarr: The reason I didn't is because I realized 1 second delay was more than enough. But adding 15 seconds to each tests would mean (15 seconds * 20 tests = 300 seconds) 5 extra minutes that is unnecessary. Specifically because if the host is busy starting up, `curl` would just wait to get a response because the request would not be completed. 

I would say we can leave it to 5sec for now. And if we see issues around this, it should be a quick update.

pragnagopa: yes. 5secs is ok.

ankitkumarr: sorry didn't see your reply. Merging this.

Issue #330, opened 2019-02-18
Empty Service Bus messages cause an error


JasonBSteele:
#### Investigative information

- Developing locally
- Core Tools version: 2.4.317

#### Repro steps

Provide the steps required to reproduce the problem:

1. Create a Function with a Service Bus Topic trigger using VS Code Azure Functions Extension
2. Update  the bindings to reference a topic subscription
3. Populate the Topic with test messages, some of which have UserProperties and an actual message payload, some of which only have UserProperties and no message payload

#### Expected behavior

The Function should be triggered successfully for both types of message

#### Actual behavior

For the messages with an empty payload the following error occurs:

```
Exception: TypeError: unable to decode incoming TypedData: unsupported combination of TypedData field None and expected binding type serviceBusTrigger
Stack:   File "d:\Workspace\MyPythonFunction\.env\lib\site-packages\azure\functions_worker\dispatcher.py", line 250, in _handle__invocation_request
    pytype=pb_type_info.pytype)
  File "d:\Workspace\MyPythonFunction\.env\lib\site-packages\azure\functions_worker\bindings\meta.py", line 291, in from_incoming_proto
    f'unable to decode incoming TypedData: '
.
```

#### Known workarounds

The workaround is to ensure all messages have a payload regardless of whether it is used or not.

#### Related information 

My Bindings:

```
{
  "scriptFile": "__init__.py",
  "disabled": false,
  "bindings": [
    {
      "name": "msg",
      "type": "serviceBusTrigger",
      "direction": "in",
      "topicName": "MyTopic",
      "subscriptionName": "MySubscription",
      "connection": "MyServiceBus_SERVICEBUS"
    }
  ]
}
```

asavaritayal:
@elprans Ping, can you look into this?

asavaritayal:
@elprans do you have an update on this?

asavaritayal:
Ping. @elprans

elprans:
@asavaritayal I'm looking into this today.

PR #371, opened 2019-04-01
Handle cases where ServiceBus message has no payload

elprans:
Apparently, a Service Bus message may have no payload at the protocol
level, so the worker must account for that.

Fixes: #330

### DIFF
--- a/azure/functions_worker/bindings/servicebus.py
+++ b/azure/functions_worker/bindings/servicebus.py
@@ -131,6 +131,11 @@ class ServiceBusMessageInConverter(meta.InConverter,
         elif data_type == 'json':
             body = data.json.encode('utf-8')
 
+        elif data_type is None:
+            # ServiceBus message with no payload are possible.
+            # See Azure/azure-functions-python-worker#330
+            body = b''
+
         else:
             raise NotImplementedError(
                 f'unsupported queue payload type: {data_type}')


azure/functions_worker/bindings/servicebus.py
```diff
@@ -131,6 +131,11 @@ def from_proto(cls, data: protos.TypedData, *,
         elif data_type == 'json':
             body = data.json.encode('utf-8')
 
+        elif data_type is None:
```
maiqbal11: Can we ever be in a situation where the data is `None` but the `data_type` is not set to `None`? I'm imagining a case where `data` takes a `None` value but the `data_type` is `string`. Is the assumption that cases like this will be handle by `grpc` itself?

elprans: We can be defensive here, but `None` value in data with `data_type` present seems to me like a functions host bug.

maiqbal11: We should be fine leaving as it is. `data_type` is assigned whenever `data` is assigned - nothing should be assigned in the case of `data` being `None`

maiqbal11:
@elprans, I believe this should be ready to merge?

maiqbal11:
Ping. I believe this should be good to check in?

PR #369, opened 2019-03-24
Read device-id from EventHub event trigger metadata

elprans:
Closes: #288

PR #375, opened 2019-04-03
v1.0.0b5

maiqbal11:


elprans:
@maiqbal11 Should we merge #369 and #371 first?

maiqbal11:
> @maiqbal11 Should we merge #369 and #371 first?

I'm fine with that. @asavaritayal had wanted to do a release, but I don't believe there are any critical fixes that we have to ship so we can wait to check those in.

marcoromelli:
When you plan to release v1.0.0.b5? I would like to test my function with the new HTTP multipart upload which is not present in v1.0.0.b4.

Issue #135, opened 2018-05-31
Do not expose protobuf to the bindings


pragnagopa:
Tracking issue to investigate decoupling azure.functions and azure.functions.worker.

Use type annotations if the function expects a custom type such as 
singnalRConnectionInfo which is not a basic type like: string or byte[] or rich binding type defined in azure-functions-python-library

```
def main(req: func.HttpRequest, connectionInfo: signalRConnectionInfo) -> func.HttpResponse:
    return func.HttpResponse(connectionInfo)
```

Python worker needs to convert the input data sent by the functions host to the type requested by the function. This is the goal of issue https://github.com/Azure/azure-functions-python-worker/issues/135 as well.

Here is the flow for the above example:
- User function includes reference to the library where `signalRConnectionInfo`  is defined
- On function invocation, host sends input data as String/Json
- Python worker examines the type annotation and converts input data to type `signalRConnectionInfo`
- If type conversion fails, throw error

asavaritayal:
Brainstorming item, not currently scheduled. @pragnagopa to update objective and problem statement.

pragnagopa:
https://github.com/Azure/azure-functions-host/issues/3140

maiqbal11:
@pragnagopa, @asavaritayal, reviving this old thread because I've got some time on my hands. Work on exposing generic data types is almost finished (https://github.com/Azure/azure-functions-python-worker/issues/66). Once that is checked in, we can potentially get rid of `azure.functions` altogether + the binding specific code in the language worker. Wanted to lay out some top-of-mind benefits and drawbacks.

Benefits:

1. Having some rich data types helps with getting users started instead of them having to deal with raw data in the form of bytes, strings etc.
2. This would be a big breaking change and would require re-education about how to use these types. 
3. It should also help with intellisense but that's a work in progress. 

Drawbacks:

1. I see the comments at Azure/azure-functions-host#3140 and it makes sense that we might not want to be in the business of owning these libraries and maintaining them. We've been bitten by this before: https://github.com/Azure/azure-functions-python-worker/issues/269. Arguably, if this had been handled by the appropriate SDK, we wouldn't have to worry about it. This adds engineering complexity as we need to maintain/service a different code base - aspects of which are not in our control if something about the format of the messages changes, for example.

A middle way might be just to leave things as they are right now and not expand the set of rich data types that we provide. We'd still maintain the ones we have but that's been okay so far. 

Lemme know what you think!

[...]

Issue #66, opened 2018-03-02
Implement base object for handling triggers with basic data types


pragnagopa:

1st1:
### Context

The original motivation for this feature is to allow creating functions for binding types that are not yet fully supported by the worker.  The *current* behaviour is to simply reject them to be loaded with an error.

The proposed solution is to:

* add common base datatypes for trigger bindings and regular bindings;
* expose the underlying `TypedData` raw data like the nodejs worker.

### Pros:

* This would mean that even if there's no direct implementation of a binding type, users would still be able to write functions that use it.  

### Cons:

* The main problem here is that by doing that, we would inadvertently expose Python Functions users to lots of **low-level** protocol details.  In the short term, users will be able to use new bindings/azure services **before** they are officially supported by the Python worker.  

* This essentially makes the low-level encoding—`TypedData` gRPC structs—**an API**.  This complicates things for binding types and azure services, as they would have to maintain backwards compatibility w.r.t. how exactly bindings are encoded on the wire.

* Moreover, we'll still have to document how exactly those bindings are encoded on the wire, otherwise users won't be able to work with these new binding types.

Therefore, implementing this feature doesn't really make it easier for users to work with unsupported bindings.  At best this would allow them to quickly prototype a function before the binding/integration is fully supported, but by doing so, they would be writing code that relies on private undocumented data formats.

--------

### Initial attempt to implement the feature

Invariants:

* We don't want to expose the actual gRPC protocol to the user.
* We don't want to mirror `TypedData` to equivalent Python classes—that's not Pythonic.
* We don't want to assume that all bindings are **always** encoded as JSON.  There might be a new binding in the future that uses `TypedData(int=..)` gRPC type for encoding its values.

Therefore we want a high-level facade types for the `TypedData`.  I've prototyped the following ABCs:

```python
class RawInputData(abc.ABC):

    @abc.abstractmethod
    def get_str(self) -> typing.Optional[str]:
        pass

    @abc.abstractmethod
    def get_bytes(self) -> typing.Optional[bytes]:
        pass

    @abc.abstractmethod
    def get_int(self) -> typing.Optional[int]:
        pass

    @abc.abstractmethod
    def get_float(self) -> typing.Optional[float]:
        pass

    @abc.abstractmethod
    def get_json(self) -> typing.Any:
        pass

    # XXX: Not complete! Probably need to add support for TypedData(http)


class RawOutputData(RawInputData):

    @abc.abstractmethod
[...]

PR #379, opened 2019-04-08
[Do not merge] Add a test for custom bindings

anthonychu:
Does the new support for raw data types allow bindings that are not known to the worker? I wasn't able to get the SignalRConnectionInfo binding working. Here's a test to show what I'm trying to do. Thanks.


pragnagopa:
cc: @maiqbal11 @asavaritayal 

@elprans  - Can you please take a look? Should the dataType be set in function.json?

elprans:
@pragnagopa Yes, for custom bindings the dataType should be set in function.json

anthonychu:
@elprans Thanks. Do you have an example?

elprans:
@anthonychu 

https://github.com/Azure/azure-functions-python-worker/blob/b31ede864d664e10388a7b0802ae162c0ea6abaa/tests/blob_functions/get_blob_as_bytes/function.json#L1-L24

anthonychu:
I tested this out and it doesn't work for `signalRConnectionInfo` because the binding returns an object and it doesn't convert to a string. I suspect other bindings that do not convert to string will have the same problems.

Is it possible for it to work like the JavaScript worker when the dataType is not specified and allow binding to a `dict` or `list`? Example:

```json
    {
      "type": "signalRConnectionInfo",
      "direction": "in",
      "name": "connectionInfo",
      "hubName": "chat"
    }
```

```javascript
def main(req: func.HttpRequest, connectionInfo: dict) -> func.HttpResponse:
      return func.HttpResponse(json.dumps(connectionInfo), mimetype='application/json')
```

elprans:
> because the binding returns an object and it doesn't convert to a string. 

Can you elaborate here?   On the protocol level everything is either a string or an array of bytes.  It seems what you are looking for is the ability to say ``dataType: "json"``, but AFAIK, the host doesn't support that.  /cc @pragnagopa

anthonychu:
I believe an undefined `dataType` is essentially the same as `dataType: "json"`. I think we should be able to handle this case for custom bindings. Here's what I'm thinking of:

If `dataType` is undefined:
* pytype is `str`: pass the JSON string
* pytype is `dict`: pass a dict if it can deserialize into one, else fail
* pytype is `list`: pass a list if it can deserialize into one, else fail
* pytype is `None`: pass the result of `json.loads()`

elprans:
Undefined `dataType` does not actually imply JSON or any structure that the worker should be able to decode.  The current handling is that an unspecified `dataType` means "use the rich binding if any".  The current design does not require the use of type annotations in user functions, so we cannot rely on them, only on what's been specified in `function.json`. There has been a lengthy discussion about this in #66.

anthonychu:
That makes sense. But even without `dataType` or type annotations, can we still perform a best effort conversion as I outlined above? That would make the Python implementation match the JavaScript one.

elprans:
Matching the JavaScript implementation is not a goal.  The Python implementation provides "rich" bindings by default, so the behavior is closer to C# rather than JavaScript.  Doing an implicit `json.loads()` without an explicitly specified `dataType` in function.json prevents of possible future implementation of a rich binding for an extension in a consistent way.  We decided in #66 (and in other discussions around it) that any form of "generic" binding must require an explicit opt-in in function.json.

pragnagopa:
@elprans - if `dataType` is set to `string` input can be a json string or just a string.  Python worker needs to check if the input is valid json first and fall back to string otherwise. Is this possible?

@anthonychu  - let me know if providing singnalRInfo as json would unblock your scenario.

pragnagopa:
cc @jeffhollan for more context

elprans:
> Python worker needs to check if the input is valid json first and fall back to string otherwise.

@pragnagopa Why can't we add an explicit "json" setting?  Implicit decoding attempt seems very fragile to me.  Changing the *type* of the value based on the input (as opposed to an explicit declaration) isn't a good thing.

Unless we declare and document otherwise, the worker cannot rely on the type annotations.  Otherwise we create a weird situation where it's not clear whether function.json is authoritative or the function declaration is.

jeffhollan:
@anthonychu I imagine your goal here is to make sure the bindings behave even if the user doesn't explicitly set a `dataType`?

@elprans is the alternative here so that a blank / default `dataType` would work to provide a "rich" binding for the SignalR stuff @anthonychu is working on?  I'm guessing a rich data type provides a bit more control for the binding to default to a desired type?

anthonychu:
I think it would be okay to set a `dataType`, but I don't think `string` would be the right one because (I think) this requires every extension used in this manner to implement string converters. If there's no converter available, as in the case of the SignalR bindings, the host will throw an exception when `dataType` is set to `string`.

If `dataType` must be explicitly set, I like @elprans's suggestion of adding a new value such as `json`. The host would treat this exactly the same if it was undefined. But the Python worker would use it to opt into running `json.loads()` on the value.

pragnagopa:
`json` is not a dataType. Functions host is built on WebJobs SDK which [deserializes input to a Json object if the input data is valid json](https://github.com/Azure/azure-webjobs-sdk/wiki/Queues#poco-plain-old-clr-object-queue-messages).  Language workers are designed to be data converters between functions host and user code.

I think right approach is to use type annotations if the function expects a custom type such as 
singnalRConnectionInfo which is not a basic type like: string or byte[] or rich binding type defined in azure-functions-python-library

```
def main(req: func.HttpRequest, connectionInfo: signalRConnectionInfo) -> func.HttpResponse:
    return func.HttpResponse(connectionInfo)
```

Python worker needs to convert the input data sent by the functions host to the type requested by the function. This is the goal of issue https://github.com/Azure/azure-functions-python-worker/issues/135 as well.

Here is the flow for the above example:
- User function includes reference to the library where `signalRConnectionInfo`  is defined
- On function invocation, host sends input data as String/Json
- Python worker examines the type annotation and converts input data to type `signalRConnectionInfo`
- If type conversion fails, throw error
**Note**: This is the first step of adding support to live objects i.e convert blob input to type '`blob` defined in [azure storage python sdk]( https://github.com/Azure/azure-storage-python/)

@elprans - Any suggestions on how above flow can be achieved?

anthonychu:
Resolved with #409

PR #389, opened 2019-04-24
Kafka Trigger Rich Type

priyaananthasankar:
Supporting rich type KafkaEvent for KafkaTrigger in Python
Azure/azure-functions-kafka-extension#10

### DIFF
--- a/azure/functions_worker/bindings/kafka.py
+++ b/azure/functions_worker/bindings/kafka.py
@@ -1,0 +1,97 @@
+import json
+import typing
+
+from azure.functions import _kafka
+
+from . import meta
+from .. import protos
+
+
+class KafkaConverter(meta.InConverter, meta.OutConverter,
+                        binding='kafka'):
+
+    @classmethod
+    def check_input_type_annotation(cls, pytype: type) -> bool:
+        return issubclass(pytype, _kafka.KafkaEvent)
+
+    @classmethod
+    def check_output_type_annotation(cls, pytype) -> bool:
+        return (
+            issubclass(pytype, (str, bytes))
+            or (issubclass(pytype, typing.List)
+                and issubclass(pytype.__args__[0], str))
+        )
+
+    @classmethod
+    def from_proto(cls, data: protos.TypedData, *,
+                   pytype: typing.Optional[type],
+                   trigger_metadata) -> _kafka.KafkaEvent:
+        data_type = data.WhichOneof('data')
+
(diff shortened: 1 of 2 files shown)


azure/functions_worker/bindings/kafka.py
```diff
@@ -0,0 +1,101 @@
...
+from . import meta
+from .. import protos
+
+
+class KafkaConverter(meta.InConverter, meta.OutConverter,
+                        binding='kafka'):
+
+    @classmethod
+    def check_input_type_annotation(
+           cls, pytype: type, datatype: protos.BindingInfo.DataType) -> bool:
```
priyaananthasankar: Added extra argument - in sync with meta.py

azure/functions_worker/bindings/kafka.py
```diff
@@ -0,0 +1,101 @@
...
+            body = data.json.encode('utf-8')
+
+        else:
+            raise NotImplementedError(
+                f'unsupported event data payload type: {data_type}')
+
+        return _kafka.KafkaEvent(
+            body=body,
+            timestamp=cls._decode_trigger_metadata_field(
+                trigger_metadata, 'Timestamp', python_type=str),
```
maiqbal11: Mentioned in https://github.com/Azure/azure-functions-python-library/pull/13 as well. Keep this as `datetime.datetime`?

priyaananthasankar: Given the different kinds of datetime formats that arrive and there is no specification on what should be a time format, I thought its best to leave it as string and allow folks to convert the way they want it.

maiqbal11:
@priyaananthasankar, looks good to me overall. Once https://github.com/Azure/azure-functions-python-worker/pull/398 is checked in and released, would be better to check this change into the Kafka extensions repo. You can include the entry point into the `setup.py` for the kafka trigger package. This would be a cleaner way of just collecting all the kafka extension related code in one place (this is a more general pattern we are recommending). \cc @pragnagopa

pragnagopa:
@maiqbal11 - Can we get a release for the python worker?
@priyaananthasankar  - As @maiqbal11 mentioned please move this PR to  https://github.com/Azure/azure-functions-kafka-extension

maiqbal11:
@pragnagopa, I'll start the release.

maiqbal11:
Implemented in https://github.com/Azure/azure-functions-kafka-extension/pull/92

PR #399, opened 2019-04-29
Fix issue with Date Time Format.

lanrongwen:
When using Queue Storage and triggering manually it would fail with this error:
```
Result: Failure
Exception: ValueError: time data '9999-12-31T23:59:59.9999999+00:00' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'
Stack:   File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/dispatcher.py", line 273, in _handle__invocation_request
    pytype=pb_type_info.pytype)
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 298, in from_incoming_proto
    trigger_metadata=trigger_metadata)
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/queue.py", line 93, in from_proto
    trigger_metadata, 'ExpirationTime'),
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 149, in _parse_datetime_metadata
    return cls._parse_datetime(datetime_str)
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 192, in _parse_datetime
    raise last_error
  File "/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 187, in _parse_datetime
    dt = datetime.datetime.strptime(datetime_str, fmt)
  File "/usr/local/lib/python3.6/_strptime.py", line 565, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/usr/local/lib/python3.6/_strptime.py", line 362, in _strptime
    (data_string, format))
```

This is to address the fact that the RegEx would miss certain formats and then wouldn't truncate correctly.

### DIFF
--- a/azure/functions_worker/bindings/meta.py
+++ b/azure/functions_worker/bindings/meta.py
@@ -167,20 +167,22 @@ class _BaseConverter(metaclass=_ConverterMeta, binding=None):
         # UTC ISO 8601 assumed
         formats = [
             '%Y-%m-%dT%H:%M:%S+00:00',
+            '%Y-%m-%dT%H:%M:%S-00:00',
             '%Y-%m-%dT%H:%M:%S.%f+00:00',
+            '%Y-%m-%dT%H:%M:%S.%f-00:00',
             '%Y-%m-%dT%H:%M:%SZ',
             '%Y-%m-%dT%H:%M:%S.%fZ',
         ]
         dt = None
 
-        too_fractional = re.match(r'.*\.\d{6}(\d+)', datetime_str)
+        too_fractional = re.match(r'(.*\.\d{6})(\d+)(Z|[\+|-]\d{1,2}:\d{1,2})', datetime_str)
         if too_fractional:
             # The supplied value contains seven digits in the
             # fractional second part, whereas Python expects
             # a maxium of six, so strip it.
             # https://github.com/Azure/azure-functions-python-worker/issues/269
-            extra_digits = len(too_fractional.group(1))
-            datetime_str = too_fractional.group(0)[:-extra_digits - 1] + 'Z'
+            extra_digits = len(too_fractional.group(2))
+            datetime_str = too_fractional.group(1)[:-extra_digits] + too_fractional.group(3)
 
         for fmt in formats:
             try:
--- a/azure/functions_worker/bindings/meta.py
+++ b/azure/functions_worker/bindings/meta.py
@@ -173,14 +173,14 @@ class _BaseConverter(metaclass=_ConverterMeta, binding=None):
         ]
         dt = None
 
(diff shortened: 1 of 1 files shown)


azure/functions_worker/bindings/meta.py
```diff
@@ -173,14 +173,14 @@ def _parse_datetime(
         ]
         dt = None
 
-        too_fractional = re.match(r'.*\.\d{6}(\d+)Z', datetime_str)
+        too_fractional = re.match(r'.*\.\d{6}(\d+)', datetime_str)
```
elprans: This would make inputs that *do* send `Z` fail, so you want `Z?` here instead.

lanrongwen: It had Z and was failing because of it, as the manual trigger is 9999-12-31T23:59:59.9999999+00:00 with no Z.  This should work with the Z as well as it should truncate from before the Z.  I can double check.

lanrongwen: Yes, in my test harness it will move the Z correctly.  Technically it strips it off then re-adds the Z on line 183

lanrongwen: Notice on 183 it will pull from the start of the string all the way until the match, so it is short of the Z, but adds one there.

elprans: Right, in this case changing `Z` to `(Z|[\+|-]\d{1,2}:\d{1,2})` is more appropriate.  We don't want so silently truncate the timezone information.

lanrongwen: Updated with code to grab the TZ and append it if it was provided.  Also, updated the format filters to include - (negative) TZ offsets.

lanrongwen: It will still add Z if that is the format given.

lanrongwen:
@elprans  Quick Q - Is there more I need to do for this PR or is it good to go?  Just checking as it's my first PR for this project, just getting the feel for it. :)  I see it's waiting on CI, but I'm not sure why that didn't run.

unstephenk:
I am getting this error on my end. How can I resolve it?

maiqbal11:
Hi @lanrongwen, apologies for the very late reply on this! Was wondering if you would still be interested in the working on this PR? We have received other reports about the DateTime format being incorrect and are looking to get in a fix very soon.

maiqbal11:
Closing in favor of https://github.com/Azure/azure-functions-python-library/pull/22 since we have moved the bindings to the library.

Issue #408, opened 2019-05-06
Custom binding error


asavaritayal:
I created an HTTP triggered function with an output binding of type Azure Table Storage. Here is the `function.json` definition: 

```json
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "authLevel": "anonymous",
      "type": "httpTrigger",
      "direction": "in",
      "name": "req",
      "methods": [
        "get",
        "post"
      ]
    },
    {
      "type": "http",
      "direction": "out",
      "name": "$return"
    },
    {
      "type": "table",
      "dataType": "string",
      "direction": "out",
      "name": "table",
      "tableName": "commitsTable",
      "connection": "AzureWebJobsStorage"
    }
  ]
}

``` 

I see the following error when trying to run locally: 

```
Executing 'Functions.PushEvent' (Reason='This function was programmatically called via the host APIs.', Id=a079f757-06a1-4ca0-b2fd-6eb41f4e8953)
[5/6/19 5:38:18 AM] Executed 'Functions.PushEvent' (Failed, Id=a079f757-06a1-4ca0-b2fd-6eb41f4e8953)
[5/6/19 5:38:18 AM] System.Private.CoreLib: Exception while executing function: Functions.PushEvent. System.Private.CoreLib: Result: Failure
[5/6/19 5:38:18 AM] Exception: FunctionLoadError: cannot load the PushEvent function: unknown type for table binding: "table"
[5/6/19 5:38:18 AM] Stack:   File "/usr/local/Cellar/azure-functions-core-tools/2.7.1149/workers/python/deps/azure/functions_worker/dispatcher.py", line 232, in _handle__function_load_request
[5/6/19 5:38:18 AM]     function_id, func, func_request.metadata)
[5/6/19 5:38:18 AM]   File "/usr/local/Cellar/azure-functions-core-tools/2.7.1149/workers/python/deps/azure/functions_worker/functions.py", line 179, in add_function
[5/6/19 5:38:18 AM]     f'unknown type for {param.name} binding: "{desc.type}"')
[5/6/19 5:38:18 AM] .
```

I expected this to work since the issue here - https://github.com/Azure/azure-functions-python-worker/issues/66 has been closed.

elprans:
@asavaritayal #66 laid the foundation for the unstructured bindings, but we didn't explicitly enable support for *arbitrary* binding types, only the bindings that are supported (registered) otherwise.  If this is desired, I can make a PR to enable this behavior.

[...]

PR #409, opened 2019-05-06
Add generic binding support

elprans:
Add a generic binding implementation to allow unknown bindings with
`dataType: string` and `dataType: binary`.

Fixes: #66
Fixes: #408


tests/generic_functions/foobar_as_str/function.json
```diff
@@ -0,0 +1,17 @@
+{
+  "scriptFile": "main.py",
+  "bindings": [
+    {
+      "type": "foobar",
```
pragnagopa: Add a test for queue trigger, with datatype set to string

azure/functions_worker/bindings/generic.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @classmethod
+    def check_input_type_annotation(
+            cls, pytype: type, datatype: protos.BindingInfo.DataType) -> bool:
+        if (datatype is protos.BindingInfo.binary):
+            return issubclass(pytype, bytes)
+        elif (datatype is protos.BindingInfo.string):
+            return issubclass(pytype, str)
+        else:  # Unknown datatype
+            return False
```
pragnagopa: Handling conversion to `Plain old Python Object` if the function signature specifies a python object type via annotation, will unblock adding new bindings that use direct types without having to create converters.  Here is an example of SingnalR binding in Java:https://github.com/Azure/azure-functions-signalrservice-extension/blob/dev/binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRConnectionInfo.java , this is a POJO

elprans: There is no such thing as a "plain old Python object", unfortunately.  `dataclasses` in Python 3.7 is the closest thing, and IMO support for that should be a separate issue.

tests/generic_functions/foobar_as_str/function.json
elprans: Done.

pragnagopa:
Please open a separate issue to add support to dataclasses with the details

tests/queue_functions/queue_trigger_untyped/main.py
```diff
@@ -0,0 +1,8 @@
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def main(msg: str) -> str:
```
pragnagopa: please add another test without annotation.  The purpose of explicity specifying dataType in function.json is to to binding to basic types without annotations

azure/functions_worker/bindings/queue.py
```diff
@@ -60,6 +60,10 @@ def check_input_type_annotation(
             cls, pytype: type, datatype: protos.BindingInfo.DataType) -> bool:
         if datatype is protos.BindingInfo.undefined:
             return issubclass(pytype, azf_abc.QueueMessage)
+        elif (datatype is protos.BindingInfo.binary):
+            return issubclass(pytype, bytes)
```
pragnagopa: what about other bindings? shouldn't they be updated to handle dataType binary and string? what about outputbindings?

elprans: Output bindings already support elementary types, where applicable.

> what about other bindings?

Which bindings do you have in mind?  Is there actually value in providing raw input for all bindings given that rich bindings already exist?  I'm thinking an abundance of options here would create confusion without actual benefit.

pragnagopa: Goal of this work item is to enable all bindings to work when `datatype` is specified in `function.json`
This means that, worker when issue #135 is addressed, users should be able to use functions without adding reference to azure.functions

elprans: This still seems to me like a step backwards.  I understand why raw interface might be useful for when there is no canonical `azure.functions` binding yet, essentially as a workaround.

But our original direction with the design of the worker was to expose well-documented Pythonic interfaces as the API, not raw, untyped, data.  Promoting the latter creates a strong binding to the internals, and that may restrict our ability to upgrade existing users to newer versions of the platform.

pragnagopa: Using azure.functions is still the recommended approach. All the documentation and the samples will be using types in azure.functions package. 

Again, this work item is to make python worker an interface layer that converts to requested types.

asavaritayal:
@elprans can you add tests here for Azure Table Storage and SignalR bindings? 

- https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-table
- https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-signalr-service

These are two of the scenarios we don't provide rich bindings for and want to make sure they're tracked and tested with each release.

pragnagopa:
With these changes, i think both table and signalr changes should work. @elprans - Please test E2E.

elprans:
@asavaritayal @pragnagopa I'll open a separate issue for SignalR and Table Storage.  This PR already has tests for arbitrary bindings.

pragnagopa:
I understand setup needed for testing signalR is difficult. Can you please add E2E test for table storage? It just needs storage account connection string which E2E tests are already using. I can signoff once table storage is tested.

elprans:
@pragnagopa How do I add Table API to an existing Cosmos DB account?

pragnagopa:
Here is the documentation for table storage bindings: https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-table for https://azure.microsoft.com/en-us/services/storage/tables/ . Please note these are not related to cosmosdb

elprans:
> Please note these are not related to cosmosdb

So is this the wrong doc? https://docs.microsoft.com/en-us/azure/cosmos-db/tutorial-develop-table-dotnet

pragnagopa:
That's correct. The doc you referred to is for CosmosDB Table API. The test I am referring to is just for Azure Table Storage. Which is similar to Blob, Queue bindings we have already.

elprans:
@pragnagopa I get this failure:

```
fail: Function.table_input[0]
      Executed 'Functions.table_input' (Failed, Id=7ad33ddf-12de-4d7b-8e54-cdc444fdfc12)
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.table_input ---> System.InvalidOperationException: Table entity types must provide a default constructor.
   at Microsoft.Azure.WebJobs.Host.Tables.TableClient.VerifyDefaultConstructor(Type entityType) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\Tables\TableClient.cs:line 75
   at Microsoft.Azure.WebJobs.Host.Tables.PocoEntityArgumentBindingProvider.TryCreate(ParameterInfo parameter) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\Tables\PocoEntityArgumentBindingProvider.cs:line 24
   at Microsoft.Azure.WebJobs.Host.Tables.CompositeEntityArgumentBindingProvider.TryCreate(ParameterInfo parameter) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\Tables\CompositeEntityArgumentBindingProvider.cs:line 23
   at Microsoft.Azure.WebJobs.Host.Tables.TableAttributeBindingProvider.TryCreate(BindingProviderContext context) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\Tables\TableAttributeBindingProvider.cs:line 69
   at Microsoft.Azure.WebJobs.Host.Tables.TableAttributeBindingProvider.<>c__DisplayClass5_0.<TryCreateAsync>b__0() in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\Tables\TableAttributeBindingProvider.cs:line 84
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Azure.WebJobs.Host.Bindings.GenericCompositeBindingProvider`1.TryCreateAsync(BindingProviderContext context) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingProviders\GenericCompositeBindingProvider.cs:line 61
   at Microsoft.Azure.WebJobs.Host.Bindings.CompositeBindingProvider.TryCreateAsync(BindingProviderContext context) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingProviders\CompositeBindingProvider.cs:line 25
   at Microsoft.Azure.WebJobs.Binder.BindAsync[TValue](Attribute[] attributes, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\Runtime\Binder.cs:line 97
   at Microsoft.Azure.WebJobs.Script.Binding.FunctionBinding.BindStringAsync(BindingContext context) in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Binding\FunctionBinding.cs:line 220
   at Microsoft.Azure.WebJobs.Script.Binding.ExtensionBinding.BindAsync(BindingContext context) in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Binding\ExtensionBinding.cs:line 75
   at Microsoft.Azure.WebJobs.Script.Description.WorkerLanguageInvoker.<>c__DisplayClass7_0.<<BindInputsAsync>b__1>d.MoveNext() in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Description\Rpc\WorkerLanguageInvoker.cs:line 95
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Azure.WebJobs.Script.Description.WorkerLanguageInvoker.BindInputsAsync(Binder binder) in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Description\Rpc\WorkerLanguageInvoker.cs:line 98
   at Microsoft.Azure.WebJobs.Script.Description.WorkerLanguageInvoker.InvokeCore(Object[] parameters, FunctionInvocationContext context) in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Description\Rpc\WorkerLanguageInvoker.cs:line 56
   at Microsoft.Azure.WebJobs.Script.Description.FunctionInvokerBase.Invoke(Object[] parameters) in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Description\FunctionInvokerBase.cs:line 84
   at Microsoft.Azure.WebJobs.Script.Description.FunctionGenerator.Coerce[T](Task`1 src) in C:\azure-webjobs-sdk-script\src\WebJobs.Script\Description\FunctionGenerator.cs:line 225
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionInvoker`2.InvokeAsync(Object instance, Object[] arguments) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionInvoker.cs:line 63
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.InvokeAsync(IFunctionInvoker invoker, ParameterHelper parameterHelper, CancellationTokenSource timeoutTokenSource, CancellationTokenSource functionCancellationTokenSource, Boolean throwOnTimeout, TimeSpan timerInterval, IFunctionInstance instance) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 556
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithWatchersAsync(IFunctionInstance instance, ParameterHelper parameterHelper, ILogger logger, CancellationTokenSource functionCancellationTokenSource) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 503
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithLoggingAsync(IFunctionInstance instance, ParameterHelper parameterHelper, IFunctionOutputDefinition outputDefinition, ILogger logger, CancellationTokenSource functionCancellationTokenSource) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 439
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithLoggingAsync(IFunctionInstance instance, FunctionStartedMessage message, FunctionInstanceLogEntry instanceLogEntry, ParameterHelper parameterHelper, ILogger logger, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 249
   --- End of inner exception stack trace ---
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithLoggingAsync(IFunctionInstance instance, FunctionStartedMessage message, FunctionInstanceLogEntry instanceLogEntry, ParameterHelper parameterHelper, ILogger logger, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 293
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.TryExecuteAsync(IFunctionInstance functionInstance, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 89
```

*function.json*:

```json
{
  "scriptFile": "__init__.py",
  "disabled": false,

  "bindings": [
    {
      "type": "httpTrigger",
      "direction": "in",
      "name": "req"
    },
    {
      "direction": "in",
      "type": "table",
      "name": "docs",
      "dataType": "string",
      "partitionKey": "test",
      "rowKey": "foo",
      "tableName": "test",
      "connection": "AzureWebJobsStorage"
    },
    {
      "type": "http",
      "direction": "out",
      "name": "$return"
    }
  ]
}
```

pragnagopa:
looks like specifying `dataType` is causing the problem. Can you try output binding?

elprans:
@pragnagopa Isn't the point here to test unstructured binding via 'dataType'?

pragnagopa:
> Isn't the point here to test unstructured binding via 'dataType'?

Goal of E2E testing
1.  Use dataType and test a trigger/input using generic binding 
2.  Test generic output binding i.e a binding that is not defined in azure.functions package

For 1, open an issue on the host. Parsing table input data as string is failing.
To test 2, try Table storage output binding

I think worker needs to handle any binding that is not defined in azure.functions package as a generic binding without enforcing the use of 'dataType` in function.json

elprans:
> I think worker needs to handle any binding that is not defined in azure.functions package as a generic binding without enforcing the use of 'dataType` in function.json

I don't think this is a good idea.  The Rich/Unstructured switch has to be explicit.  Otherwise, if we choose to support a binding in a future release, user code will break.

pragnagopa:
> I don't think this is a good idea. The Rich/Unstructured switch has to be explicit. Otherwise, if we choose to support a binding in a future release, user code will break.

Can you explain? Generic binding is only fall back if there are no rich types found in any of the referenced assemblies

elprans:
I cannot write a decent E2E test if only half of the host functionality is working.  How do I test that the output binding is working correctly without being able to *read* the data back?

pragnagopa:
> I cannot write a decent E2E test if only half of the host functionality is working. How do I test that the output binding is working correctly without being able to read the data back?

You do not have to read data to write data to an output binding. Let's take one step at a time. Does writing using generic output binding work?

elprans:
> You do not have to read data to write data to an output binding. Let's take one step at a time. Does writing using generic output binding work?

The function doesn't crash, but I have no way to verify if the data was actually written.

pragnagopa:
> The function doesn't crash, but I have no way to verify if the data was actually written.

Ok. I understand you meant you cannot automate. For now, connect to azure storage using https://azure.microsoft.com/en-us/features/storage-explorer/ and verify.

Next, 
1. Can you explain the downside for falling back to generic bindings without dataType? 
2. Go ahead and open an issue on the host about table binding and dataType set to String does not work.

elprans:
> Can you explain the downside for falling back to generic bindings without dataType?

Scenario:

1. User installs worker version X, writes a bunch of functions without `dataType` for the Table binding. Their functions receive raw JSON.

2. We decide we want to support Table bindings natively.  Worker version X+1 released.

3. User upgrades to worker version X+1.  All their Table functions break, since we are now returning rich Table objects instead of JSON.

elprans:
For the input crash: https://github.com/Azure/azure-functions-host/issues/4442

pragnagopa:
Issue #135 ensures that types are loaded from azure.functions package and not the worker. This also ensures that user provides the version of azure.functions package they want to use in their functions.

This addresses your concern (which is valid) about breaking changes. Any breaking changes will be shipped as major versions of the azure.functions package.

elprans:
The fact that users will be upgrading `azure.functions` makes this only marginally better.  We *can* detect a missing `dataType` declaration and signal it clearly in the case where there is an unsupported binding.  On the other hand, without an explicit `dataType` it is impossible to understand the intent.  Basically we are trading the very small inconvenience of specifying `dataType` for a massive PITA later when they decide to upgrade.

pragnagopa:
Letting users upgrade `azure.functions` or any other types that follow binding model you have described here: https://github.com/Azure/azure-functions-python-worker/wiki/Adding-support-for-a-new-binding-type gives them more control over dependencies. 

I do not think relying on dataType to figure out  type by user requested is the right thing to do specifically when dataType  defaults to string. Worker has all the information to infer the type. Here is the order of precedence
1. A binding type that follows extension model defined in any package
2. Types from azure.functions
3. Fall back to generic binding.

@maiqbal11 / @asavaritayal  - Please chime in if you disagree.

elprans:
We should at least consider making type annotations required for rich bindings.

pragnagopa:
Yes. Requiring type annotations makes this change a deterministic approach. But my understanding is annotations are completely optional and we cannot impose a restriction on users to specify annotations.

@asavaritayal what do you think?

pragnagopa:
discussed offline with rest of the team.  Please update the PR that follows the order of precedence mentioned above. For any bindings that are not found in custom binding packages or azure.functions, fallback to generic bindings.

Let me know if you have any other questions.

asavaritayal:
> Yes. Requiring type annotations makes this change a deterministic approach. But my understanding is annotations are completely optional and we cannot impose a restriction on users to specify annotations.
> @asavaritayal what do you think?

Correct, annotations are an optional component in Python. We don't expect all Functions users to be pro-developers and hence cannot put the additional constraint on them of having to provide suitabel types for their bindings. Please proceed with the order of precedence recommended above by Pragna.

pragnagopa:
Ping!

elprans:
@pragnagopa @asavaritayal I still think doing this is a bad idea, but, assuming the risks of going with this approach have been discussed and accepted, I'll proceed with this change.

pragnagopa:
cc: @anirudhgarg 
@elprans  - After another round of discussions, we still agree that inferring types from the imported package at the function app level is the right approach. For dependencies, Function App level is the boundary and not each function.

Please update the PR:
- Remove dependency on `dataType`
- If azure.functions is imported, for any binding defined in `azure.functions` package, user code will binding to types defined in `azure.functions`
- If azure.functions is not imported, then fall back to generic bindings

anirudhgarg:
Yes wanted to further provide context here. We discussed this in detail and we compared the two proposals - Forcing to have explicit annotations v/s *For those types that are part of azure.functions* based on whether import azure.functions is added in the Function use the rich bindings v/s the generic bindings. 

Based on these choices in order to:
-Not break a large number of existing apps
-The scenario of somebody wanting to use generic bindings for those types that are part of azure.functions in a function *and* at the same time use typed bindings in another function with the same app should be very very rare; 

we decided that we should do the following as Pragna pointed out: 

- Remove dependency on dataType
- If azure.functions is imported, for any binding defined in azure.functions package, user code will binding to types defined in azure.functions
- If azure.functions is not imported, then fall back to generic bindings
-

elprans:
@anirudhgarg I'm assuming the scenario I described in https://github.com/Azure/azure-functions-python-worker/pull/409#issuecomment-492841454 is not considered a likely problem?

pragnagopa:
@elprans we do not plan on expanding current rich types defined in azure.functions
We will start investing in live objects support

anirudhgarg:
@elprans This is a great point and @pragnagopa mentioned what we are planning to do. I will expand more on that. 

You are essentially asking how would we do versioning. This is our plan. The current azure.functions implementation will not be changed and will be the default baked in the image. New versions of azure.functions etc. will be tied to an extension bundle version whose version the customer can control. We will change the order in which we load the azure.functions module where we will always load the extension bundle version first and then the default azure.functions version. 

Now if a customer has some existing code and it is working fine with the current default azure.functions module he can just continue to use it. If he wants to use a new version of the azure.functions module which is part of an extensions bundle the he is *explicitly* doing that and if that has any breaking changes then he should be able to do that explicitly. So in your example if rich types are introduced later for Table storage then it will be done in an azure.functions module along with an extension bundle. If the customer does not use it he should be all good. When he does use it, he will have to update his code.

elprans:
There is another thing I forgot to mention.  Magic imports are generally considered non-Pythonic, so flake8 will report an "unused import" error on `import azure.functions`.

>  We will start investing in live objects support
What are live objects?

asavaritayal:
@elprans that's okay. It's up to the user on whether to leverage the library for annotations in their code or not. I can go over live objects later. Request you to not block on that for now.

pragnagopa:
> What are live objects?

We will start working on the design for how to use pass a data structure that is same as sdk object for  the bindings we support via rpc layer to language workers. For example, for eventhub trigger, host will send all the information needed to create live object of type: `eventhubevent` defined in azure-sdk-for-python. 

As @asavaritayal mentioned, please do not block on updating this PR. This is a needed change even after live objects support. At any point of time, generic bindings are a fallback if there is no rich type or live object for a binding

elprans:
> We will start working on the design for how to use pass a data structure that is same as sdk object for the bindings we support via rpc layer to language workers.

Thanks for the explanation.  That sounds like a good plan.

pragnagopa:
Ping!

elprans:
@pragnagopa I'm working on it.  The magic import functionality requires moving all current bindings to `azure-functions-python-library` that's a bit of work.

pragnagopa:
If it helps please  split this into separate issues/PRs

elprans:
@pragnagopa Certainly.

elprans:
For consideration: leaving out both the `dataType` and the function annotations makes it very hard to reason about the actual type of the data received by the binding.  We cannot simply say that the default is 'string', because there are bindings where that makes no sense (i.e. HTTP and Blob).  This is not a blocker per-se, just a point to remember that every extension would need to document its default data format very prominently.

pragnagopa:
`dataType`  is handled by the host. Except for httpTrigger, rest of the bindings default to string.  I agree we can improve documentation here: https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings to add details on httpTrigger. 
Language worker  receives data as bytes if `dataType` is set to binary.  Python worker should not convert data to string if `dataType` is set `binary`

elprans:
@pragnagopa What about the Blob binding?  It seems to be "binary" by default.  `EventHub` events may also contain [binary data](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.eventdata.-ctor?view=azure-dotnet#Microsoft_ServiceBus_Messaging_EventData__ctor_System_Byte___).

elprans:
The `Queue` messages may also be [binary](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.queue.cloudqueuemessage.-ctor?view=azure-dotnet#Microsoft_Azure_Storage_Queue_CloudQueueMessage__ctor_System_Byte___)

pragnagopa:
Correcting my earlier comment
> Language worker receives data as bytes if dataType is set to binary. Python worker should not convert data to string if dataType is set binary

Correction:
For generic bindings, when python worker receives data as bytes, it should not convert to any other type unless requested by the function signature via annotations. 
`dataType` is used by the host to either send bytes or string. Language worker should be not be responsible for type conversion irrespective of bindings. This is specifically needed to avoid any knowledge of bindings by the language worker.

elprans:
@pragnagopa So you are saying the worker should ignore the `dataType` altogether and rely only on the presence of the magic `azure.functions` import or explicit function annotations?

elprans:
@pragnagopa Also, how does all this apply to the output bindings?  How does the worker determine the correct output format for a generic binding?  Is this the responsibility of a user then?

pragnagopa:
> So you are saying the worker should ignore the dataType altogether and rely only on the presence of the magic azure.functions import or explicit function annotations?

that's correct!

>  Also, how does all this apply to the output bindings? How does the worker determine the correct output format for a generic binding? Is this the responsibility of a user then?

Yes. Output bindings support basic types and plain old objects (json). Functions host does not have any built in type conversion for output bindings.  Here is an example of javascript function - blobtrigger in and out, blob output binding string and blob output binding json

```js
module.exports = async function (context, myBlob) {
    context.log("JavaScript blob trigger function processed blob \n Name:", context.bindingData.name, "\n Blob Size:", myBlob.length, "Bytes");
    context.bindings.myOutputBlob = context.bindings.myInputBlob;
    context.bindings.myOutputBlobString = "stringout";
    var text = '{ "employees" : [' +
'{ "firstName":"John" , "lastName":"Doe" },' +
'{ "firstName":"Anna" , "lastName":"Smith" },' +
'{ "firstName":"Peter" , "lastName":"Jones" } ]}';
context.bindings.myOutputBlobJson = text;
};
```

Corresponding funciton.json
```
{
  "bindings": [
    {
      "name": "myInputBlob",
      "type": "blobTrigger",
      "direction": "in",
      "path": "samples-workitems/{name}",
      "connection": "AzureWebJobsStorage"
    },
    {
      "type": "blob",
      "name": "myOutputBlobString",
      "path": "outcontainer/stringOut",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    },
    {
      "type": "blob",
      "name": "myOutputBlob",
      "path": "outcontainer/{rand-guid}",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    },
    {
      "type": "blob",
      "name": "myOutputBlobJson",
      "path": "outcontainer/jsonout",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    }
  ]
}
```
Content of the blob is set appropriately.

elprans:
> Output bindings support basic types and plain old objects

I can write a non-text blob with the current implementation.  Are you saying we should remove this capability?

pragnagopa:
can you give example for non-text blob?

elprans:
> can you give example for non-text blob?

I upload a PNG image or a ZIP file using an HTTP trigger, and store that as a blob.

pragnagopa:
Not sure I understand the issue you are pointing. When using a generic binding, uploading a PNG image or a ZIP file should still work. User needs to set the data on the output binding. Here is the expanded sample of java script function that uploads a local file
```js
const fs = require('fs');
const util = require('util');
const readFileAsync = util.promisify(fs.readFile);
module.exports = async function (context, myBlob) {
    context.log("JavaScript blob trigger function processed blob \n Name:", context.bindingData.name, "\n Blob Size:", myBlob.length, "Bytes");
    context.bindings.myOutputBlob = context.bindings.myInputBlob;
    context.bindings.myOutputBlobString = "stringout";
    var text = '{ "employees" : [' +
                '{ "firstName":"John" , "lastName":"Doe" },' +
                '{ "firstName":"Anna" , "lastName":"Smith" },' +
                '{ "firstName":"Peter" , "lastName":"Jones" } ]}';
    context.bindings.myOutputBlobJson = text;
    var data =  await readFileAsync('./BlobTrigger1/JavaVersion.png');
    context.log(data.length);
    context.bindings.myOutputBlobPng = data;       
};
```

corresponding function.json

```
{
  "bindings": [
    {
      "name": "myInputBlob",
      "type": "blobTrigger",
      "direction": "in",
      "path": "samples-workitems/{name}",
      "connection": "AzureWebJobsStorage"
    },
    {
      "type": "blob",
      "name": "myOutputBlobString",
      "path": "outcontainer/stringOut",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    },
    {
      "type": "blob",
      "name": "myOutputBlob",
      "path": "outcontainer/{rand-guid}",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    },
    {
      "type": "blob",
      "name": "myOutputBlobJson",
      "path": "outcontainer/jsonout",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    },
    {
      "type": "blob",
      "name": "myOutputBlobPng",
      "path": "outcontainer/{rand-guid}.png",
      "connection": "AzureWebJobsStorage",
      "direction": "out"
    }
  ]
}
```

It would be best if you can share python function code and function.json for cases you think might not work.

elprans:
OK, so in generic binding mode the function must return either a `str` instance or an object supporting the buffer protocol (`bytes`, `bytearray` etc).  If the function returns anything other than this, we raise an error.  Additionally, if the output type annotation does not match the returned value type, we *also* raise an error.  Does this sound right?

pragnagopa:
>  so in generic binding mode the function must return either a str instance or an object supporting the buffer protocol (bytes, bytearray etc). 

As long as return type matches supported types by the [TypedData defined in protobuf](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/dev/src/proto/FunctionRpc.proto#L282-L290), it should work

> if the output type annotation does not match the returned value type, we also raise an error. Does this sound right?

Yes. If user specified annotation, we should raise error if return type does not match annotation type.

pragnagopa:
Expanding on the earlier question:
> if the output type annotation does not match the returned value type, we also raise an error. Does this sound right?

Type checking only matters when using `$return` . A function can have multiple output bindings and no type restriction checks are needed when explicitly setting the value for an output binding.

pragnagopa:
Ping!

pragnagopa:
@elprans - Ping! Please give an update here or over email.

azure/functions_worker/bindings/__init__.py
```diff
@@ -1,25 +1,14 @@
...
-from . import http  # NoQA
-from . import queue  # NoQA
-from . import servicebus  # NoQA
-from . import timer  # NoQA
-
 
 __all__ = (
     'Out', 'Context',
-    'is_binding', 'is_trigger_binding',
+    'is_trigger_binding',
```
pragnagopa: why do you need 'is_trigger_binding'?

azure/functions_worker/bindings/datumdef.py
```diff
@@ -0,0 +1,82 @@
+from .. import protos
+
```
pragnagopa: Nit: Can you rename this class to rpcConverters as this is a utility to convert To and From TypedData?

azure/functions_worker/bindings/generic.py
```diff
@@ -0,0 +1,46 @@
...
+            return datumdef.Datum(type='bytes', value=bytes(obj))
+
+        else:
+            raise NotImplementedError
+
+    @classmethod
+    def decode(cls, data: datumdef.Datum, *, trigger_metadata) -> typing.Any:
+        data_type = data.type
+
+        if data_type == 'string':
```
pragnagopa: if `data_type` is `undefined` it is string

tests/queue_functions/queue_trigger_untyped/function.json
```diff
@@ -0,0 +1,22 @@
...
+  "disabled": false,
+
+  "bindings": [
+    {
+      "type": "queueTrigger",
+      "direction": "in",
+      "name": "msg",
+      "queueName": "testqueue-untyped-return",
+      "connection": "AzureWebJobsStorage",
+      "dataType": "string"
```
pragnagopa: remove dataType: string

setup.py
```diff
@@ -226,10 +226,10 @@ def run(self):
     install_requires=[
         'grpcio~=1.20.1',
         'grpcio-tools~=1.20.1',
-        'azure-functions==1.0.0b4',
     ],
     extras_require={
         'dev': [
+            'azure-functions==1.0.0b4',
```
pragnagopa: is this only needed for tests?

azure/functions_worker/bindings/datumdef.py
elprans: Not really.  The `Datum` class represents a piece of data.  It can convert to and from `TypedData`, but that's not the only way it gets used.

azure/functions_worker/bindings/generic.py
elprans: This is different from RPC `dataType`.  `Datum.type` is always defined.

setup.py
elprans: Correct.

azure/functions_worker/bindings/generic.py
pragnagopa: Is `dataType. Datum.type` set to  `string` default?

elprans: Worker ignores `dataType` completely and only pays attention to whats in `TypedData`.

tests/queue_functions/queue_trigger_untyped/main.py
maiqbal11: I believe this should no longer be applicable since we are removing the dependency on data type?

pragnagopa: We do need to have tests for both i.e dataType is specified and not specified.

tests/test_queue_functions.py
```diff
@@ -54,6 +54,18 @@ def test_queue_message_object_return(self):
         self.assertEqual(r.status_code, 200)
         self.assertEqual(r.text, 'test-message-object-return')
 
+    def test_queue_untyped_return(self):
```
maiqbal11: Not fully understanding the flow of this test case. The `put_queue_untyped_return` function adds a message to the queue `testqueue-untyped-return` via an http post. After the sleep, the `get_queue_untyped_blob_return` reads from an input blob and returns an http response. The second function call does not use the queue message from the first function - is that intended? The naming and structure of the function suggests a relation between the two.

elprans: Each test is independent to avoid races.  The flow in each trigger test is "put data" -> "wait for trigger" (trigger writes a blob) -> "check the blob to ensure trigger ran".

azure/functions_worker/bindings/__init__.py
elprans: It is used to distinguish regular input bindings from trigger bindings, because there are differences in handling these (metadata, etc.)

tests/test_queue_functions.py
maiqbal11: I'm not able to match up the code with flow that you've described. Here's what I see as the flow for each of these requests. 

1. `put_queue_untyped_return`: POST Http data -> return queue message with bytes type. This is meant to check that we are able to write queue message with `bytes`?
2. `get_queue_untyped_blob_return`: GET request -> read blob data -> return data as string via HTTP.  The function does not make use of queues so I'm not sure why queue is mentioned in the name. The intention here is to check that HTTP messages can be returned with `str` type? 

I'm also not sure why both of these calls are in the same test function `test_queue_untyped_return` even though they are completely independent and are not using the same resources either.

elprans: > The function does not make use of queues so I'm not sure why queue is mentioned in the name. The intention here is to check that HTTP messages can be returned with str type?

The intent is to verify that the queue trigger has fired.  There is no way to do this other than make the trigger produce a verifiable side-effect.  Writing blobs seems like a most reliable way.  The name of the function contains `queue` because it belongs logically to the same test.

maiqbal11:
@elprans, should be good to check in?

PR #439, opened 2019-05-29
Remove grpc tools from package

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/392


maiqbal11:
There's some really bizarre CI failure happening for `mypy` code quality checks in `3.7`. Tried repro-ing it with the same bits in my private branch using DevOps, and locally as well, but did not see it reoccur. I'm checking this in because there might be something on DevOps side causing the failure on this particular PR. Will debug some more if issue continues.

Issue #380, opened 2019-03-20
Function get_body() or req.body doesn't seem to return the actual HTTP request body


christian-vorhemus:
I have a simple HTTP triggered function which should receive a binary image and upload it to Azure Storage. The upload process fails because the HTTP request body seems to be malformed.

#### Investigative information

azure-common==1.1.18
azure-functions==1.0.0b3
azure-functions-worker==1.0.0b4

#### Repro steps

1) Create a Python Function App with the following code (note: same behaviour on JavaScript/TypeScript apps):

```
import logging
import azure.functions as func
from azure.storage.blob import BlockBlobService, PublicAccess
import sys

def main(req: func.HttpRequest, outputblob: func.Out[func.InputStream]) -> func.HttpResponse:
    logging.info('endpoint function processed a request.')

    command = req.route_params.get('command')
    file_binary = req.get_body()
    sz = sys.getsizeof(file_binary)

    outputblob.set(file_binary)
    return func.HttpResponse("Success", status_code=200)
```

2) Make a POST request to this function with a binary image in the body (Content-Type: image/png)

![code2](https://user-images.githubusercontent.com/3229424/54722737-8d8b2f80-4b65-11e9-980e-ad5dd0e5caae.png)

3) req.get_body() returns a binary object which is larger than it should be. This is what Fiddler tells me:

![code3](https://user-images.githubusercontent.com/3229424/54722744-93811080-4b65-11e9-8c68-8704eaba579c.png)

And this is what I see when debugging:

![code1](https://user-images.githubusercontent.com/3229424/54722746-97149780-4b65-11e9-9b86-047766d08f3d.png)

The file is ~635KB large and increased to ~1.1MB when accessing req.get_body()

#### Expected behavior

req.get_body() should return the raw binary HTTP request body

#### Related information 

Possibly related to #3875 - but this is closed?

jimbobbennett:
@maiqbal11 - there is a reproduction at https://github.com/jimbobbennett/PythonFunctionsBodySizeBugRepo

If I upload binary file data the data that arrives is different to the data sent - the size of the body of the request is larger and the data is different.

If you launch the function app in that repo, then run the app.py file in the App folder you will see the bug in action.

I am using:
Azure Functions Core Tools (2.5.553 Commit hash: b63f2d6f5479718ca276f7b585a8cc6b9bc57c4f)
Function Runtime Version: 2.0.12382.0

maiqbal11:
[...]

PR #453, opened 2019-06-06
Request raw http body bytes

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/380


azure/functions_worker/dispatcher.py
```diff
@@ -209,9 +209,14 @@ def _serialize_exception(self, exc):
     async def _handle__worker_init_request(self, req):
         logger.info('Received WorkerInitRequest, request ID %s',
                     self.request_id)
+
+        capabilities = dict()
+        capabilities["RawHttpBodyBytes"] = "true"
```
pragnagopa: `RawHttpBodyBytes` move this to constants. Overtime, there might be more capabilities

tests/test_http_functions.py
```diff
@@ -189,3 +190,12 @@ def test_return_route_params(self):
...
         self.assertEqual(resp, {'param1': 'foo', 'param2': 'bar'})
+
+    def test_raw_body_bytes(self):
+        image_file = pathlib.Path(__file__).parent / 'resources/functions.png'
+        with open(image_file, 'rb') as image:
+            img = image.read()
+            img_len = len(img)
+            r = self.webhost.request('POST', 'raw_body_bytes', data=img)
+        received_body_len = int(r.headers['body-len'])
+        self.assertEqual(received_body_len, img_len)
```
pragnagopa: also verify content is the same to ensure data is not lost.

PR #468, opened 2019-06-22
Null coalescing with `or` should be faster.

t-eckert:
Most minor of refactorings.


t-eckert:
@asavaritayal Any thoughts on this tiny change?

elprans:
Coalescing isn't usually faster than an `if` statement, and I usually prefer not to use it for long statements such as these.

t-eckert:
Cool! I learned something 🙂

I’ll abandon the PR

PR #472, opened 2019-06-28
Add .vscode in .gitignore

Hazhzeng:
When creating the work space in VSCode, it will automatically create a .vscode folder to store the user's setting (remote debug ptvsd etc.).

We may want to keep this folder local to each developers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ ENV/
 
 .testconfig
 .pytest_cache
+
+# vscode
+.vscode


t-eckert:
One consideration here is whether or not we want to ignore the whole directory. There may be settings that are worthwhile to keep in the repo - for instance not auto-formatting on save. I keep this off when working in the repo so I don't accidentally  reformat the entire project.

Hazhzeng:
> One consideration here is whether or not we want to ignore the whole directory. There may be settings that are worthwhile to keep in the repo - for instance not auto-formatting on save. I keep this off when working in the repo so I don't accidentally reformat the entire project.

Hmm, but since we don't have .vscode so far, we are not sharing any vscode setting.
If we need to share the vscode setting, we can just use `git add --force .vscode` to overwrite the .gitignore setting.

PR #473, opened 2019-06-28
Clear sys.path cache to import from new folders

maiqbal11:
We need to clear `sys.path` cache to allow for reloading of modules from directories that are created after we start the host (i.e. during specialization).

### DIFF
--- a/tests/path_import/test_path_import.sh
+++ b/tests/path_import/test_path_import.sh
@@ -1,0 +1,14 @@
+#! /bin/bash
+
+# $2 is sys.path from caller
+export PYTHONPATH="test_module_dir:$2"
+SCRIPT_DIR="$(dirname $0)"
+
+if [ "$1" = "fail" ]
+    then
+        python $SCRIPT_DIR/path_import.py 'fail'
+else
+    python $SCRIPT_DIR/path_import.py 'success'
+fi
+
+unset PYTHONPATH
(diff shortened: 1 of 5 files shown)


tests/test_rpc_messages.py
```diff
@@ -1,28 +1,27 @@
...
 from azure.functions_worker import testutils
 
 
 class TestGRPC(testutils.AsyncTestCase):
     pre_test_env = os.environ.copy()
 
-    async def _handle_environment_reload_request(self, test_env):
+    def _reset_environ(self):
+        for key, value in self.pre_test_env.items():
+            os.environ[key] = value
```
Hazhzeng: You may also want to unset the dangling os.environ
Since you don't have any so far, this is not a big deal.

tests/path_import/test_path_import.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+
+# $2 is sys.path from caller
+export PYTHONPATH="test_module_dir:$2"
+SCRIPT_DIR="$(dirname $0)"
+
+if [ "$1" = "fail" ]
```
Hazhzeng: Guess you can simplify this with `python $SCRIPT_DIR/path_import.py "$1"`

maiqbal11: Definitely. Will do!

PR #479, opened 2019-07-05
Copy __init__.py for azure + install azure-functions

maiqbal11:
Make some additions to populating dependencies for tooling:

1. Add `__init__.py` to `azure` module to make it a namespace
2. Install `azure.functions` separately since we no longer include it as a worker dependency


maiqbal11:
@ankitkumarr, verified that the NuGet is now produced with `__init__.py` in the `azure` module. Let me know if anything more is needed.

Issue #273, opened 2019-01-03
Noisy Error Messages for Import Module Failure


asavaritayal:
_From @raahmed on January 3, 2019 21:35_

Context: Running Python on Functions in Linux (Preview)
Severity:
Medium( Time Confusing - causes Developer Frustration)

Reproduction Steps:
0.  Add an 'import azure.storage' to your code - without installing azure.storage locally.
1. Execute 'func host start'.
2. Notice that Functions error messages are quite noisy - which makes it difficult to find relevant information. On occasion, multiple pages worth of errors have to be scrolled up to find what is needed. 
![image](https://user-images.githubusercontent.com/25991359/50662648-05c95800-0f5c-11e9-8899-056d0c05e724.png)
3. After multiple pages worth of scrolling, the true cause is found:
![image](https://user-images.githubusercontent.com/25991359/50662661-1083ed00-0f5c-11e9-9c94-0367bcc698ce.png)

Mitigation Steps:
0. Reduce the noise in the error messages so that it is easier to find the error. 

_Copied from original issue: Azure/Azure-Functions#1090_

fabiocav:
@asavaritayal assigning this to Triaged for review today.

jeffhollan:
Guessing since this is popping up on the Day 1 experience stuff this should be a P1.  Potentially a P0.  Wondering if we should also have a day-1 label as that's been an important workstream across languages.

asavaritayal:
Moving to 52 based on discussion during standup.

asavaritayal:
@ankitkumarr can help with the investigation.

Hazhzeng:
Thanks @ankitkumarr, we can modify the [traceback ](https://docs.python.org/3.7/library/traceback.html#traceback.print_tb) setting.

A quick and dirty fix to shorten the exception is to limit the stack trace in `dispatcher.py:190` (e.g. `stack_trace = ''.join(traceback.format_tb(exc.__traceback__, 2))`)

As we discussed, there are a few possible ways to allow our users to modify exception serializer:
1. add a setting option for the stack trace limit
2. allow users to pass in an exception handler

asavaritayal:
Scoping issue down to import module error, per discussion in the scrum. @elprans may be able to advise.

elprans:
For reference, Python uses this when rendering tracebacks natively:

https://github.com/python/cpython/blob/0f4e8132820947d93eccf31b9e526b81c6ffa53d/Python/import.c#L1484

[...]

PR #480, opened 2019-07-08
Add marshall exception trace for ModuleNotFoundError

Hazhzeng:
### Related Issue
Fixes https://github.com/Azure/azure-functions-python-worker/issues/273

### Solution
I implement an stackframe filter to suppress the error message from python "importlib":
1. The exception traceback will be extracted into a [StackSummary](https://docs.python.org/3/library/traceback.html#traceback.StackSummary)
2. Filter out unnecessary **<frozen importlib._bootstrap>** and **<frozen importlib._bootstrap_external>** according to @elprans cpython [reference](https://github.com/python/cpython/blob/0f4e8132820947d93eccf31b9e526b81c6ffa53d/Python/import.c#L1484)
3. Format the modified StackSummary

### CR Request
@elprans @maiqbal11


maiqbal11:
@Hazhzeng, could you create a new branch within the worker repo for the CI to run? Forks are currently not enabled for CI. Let me know if you need permissions.

tests/test_broken_functions.py
```diff
@@ -1,5 +1,6 @@
 from azure.functions_worker import protos
 from azure.functions_worker import testutils
+import traceback
```
maiqbal11: nit: Move to the top of the file as standard library imports come before custom imports by convention (https://www.python.org/dev/peps/pep-0008/#imports)

Hazhzeng:
Move to https://github.com/Azure/azure-functions-python-worker/pull/482

PR #491, opened 2019-07-19
Fix nested namespace module import issue

Hazhzeng:
This is related to running a function app with nested namespace library (e.g. azure.storage)
Repository is [here](https://github.com/jeffhollan/google-tensorflow-sample).

When we specialize the python worker from placeholder mode, we need to
1. invalidate the finder's cache.
2. reload the module to resolve the second level namespace.

Otherwise, the module will not be recognized.

### DIFF
--- a/azure/functions_worker/dispatcher.py
+++ b/azure/functions_worker/dispatcher.py
@@ -375,9 +375,10 @@ class Dispatcher(metaclass=DispatcherMeta):
             for var in env_vars:
                 os.environ[var] = env_vars[var]
 
-            if sys.modules.get('azure'):
-                importlib.reload(sys.modules['azure'])
-                logger.info('Reloading azure module')
+            importlib.invalidate_caches()
+
+            for module in sys.modules.values():
+                importlib.reload(module)
 
             success_response = protos.FunctionEnvironmentReloadResponse(
                 result=protos.StatusResult(
--- a/azure/functions_worker/dispatcher.py
+++ b/azure/functions_worker/dispatcher.py
@@ -375,10 +375,8 @@ class Dispatcher(metaclass=DispatcherMeta):
             for var in env_vars:
                 os.environ[var] = env_vars[var]
 
-            importlib.invalidate_caches()
-
-            if sys.modules.get('azure'):
-                importlib.reload(sys.modules['azure'])
+            # Reload azure namespace for customer's libraries
+            importlib.reload(sys.modules['azure'])
 
             success_response = protos.FunctionEnvironmentReloadResponse(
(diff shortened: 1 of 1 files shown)


azure/functions_worker/dispatcher.py
```diff
@@ -374,6 +375,11 @@ def _serialize_exception(self, exc):
             for var in env_vars:
                 os.environ[var] = env_vars[var]
 
+            importlib.invalidate_caches()
+
+            for module in sys.modules.values():
+                importlib.reload(module)
```
elprans: `reload()` is not safe in Python, especially when done to arbitrary modules.  Why is this needed?

Hazhzeng: @elprans Hmm, because we are facing an issue when a customer uses **azure.storage.blob** library (ModuleNotFound).

In that library, there is a nested namespace, the first level is azure (which shares the same namespace as azure.functions), while the second level is azure.storage ([here](https://github.com/Azure/azure-storage-python/tree/master/azure-storage-blob/azure/storage)). The user libraries are placed in `/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure`

When we do specialization in placeholder mode, the second level namespace "azure.storage" cannot be found.

The reason of this is when we first import **azure.functions** in placeholder mode, the **azure.__path__** is cached `['/usr/local/lib/python3.6/site-packages/azure']`.

We want to refresh the **azure** namespace to make it include the user's library location as well. So the expected **azure.__path__** should be `['/usr/local/lib/python3.6/site-packages/azure', '/home/site/wwwroot/.python_packages/lib/python3.6/site-packages/azure']`.

If there are better suggestions, please advise.
Much appreciated.

elprans: You only need to reload `azure` then: `importlib.reload(sys.modules['azure'])`

azure/functions_worker/dispatcher.py
```diff
@@ -374,6 +375,11 @@ def _serialize_exception(self, exc):
             for var in env_vars:
                 os.environ[var] = env_vars[var]
 
+            importlib.invalidate_caches()
```
elprans: Please add a comment on why this is here (e.g to repopulate `azure.__path__` on worker specialization).

Hazhzeng: Since we are only specializing the azure package, I don't think this line of code is necessary anymore. Thanks for picking this out.

Hazhzeng: Moving to the main repo (Fixed nested namespace library import issue #492)

PR #492, opened 2019-07-19
Fixed nested namespace library import issue

Hazhzeng:
Migrated from external fork: https://github.com/Azure/azure-functions-python-worker/pull/491

This is related to running a function app with nested namespace library (e.g. azure.storage)
Repository is [here](https://github.com/jeffhollan/google-tensorflow-sample).

When we specialize the python worker from placeholder mode, we need to
1. reload the azure module to resolve the second level namespace.

This issue currently impacts the customer who are using libraries within azure namespace.


tests/unittests/test_rpc_messages.py
```diff
@@ -69,3 +69,30 @@ def test_successful_sys_path_import(self):
...
+        try:
+            path_import_script = os.path.join(
+                testutils.UNIT_TESTS_ROOT,
+                'azure_namespace_import',
+                'test_azure_namespace_import.sh')
+
+            subprocess.run(['chmod +x ' + path_import_script], shell=True)
+
+            output = subprocess.check_output(
+                [path_import_script, result, 'namespace_location_b'],
```
maiqbal11: Do we also want this to be something like `$SCRIPT_DIR/namespace_location_b` so we don't have to rely on relative paths?

tests/unittests/azure_namespace_import/azure_namespace_import.py
```diff
@@ -0,0 +1,50 @@
...
+        request_id='0',
+        function_environment_reload_request=request)
+
+    disp = testutils.create_dummy_dispatcher()
+
+    # Mock intepreter starts in placeholder mode
+    import azure.module_a as mod_a  # noqa: F401
+
+    # Mock function specialization, load customer's libraries and functionapps
+    test_root = sys.argv[2]
```
maiqbal11: Can we rename this to `ns_root` or something to indicate that it is the directory with the second namespace? `test_root` is a term that we already use heavily and can mean a number of things.

Issue #504, opened 2019-07-24
Time issue: Exception: AttributeError: module 'azure.functions' has no attribute 'get_binding_registry'


docgregt:
Firstly, this code has been running for months. Secondly, the last time I deployed two days ago the deployment worked fine. Now today I deploy my function app that two timers in it and neither timer will work.
I get following errors:
Result: Failure
Exception: AttributeError: module 'azure.functions' has no attribute 'get_binding_registry'
Stack:   File "/usr/local/lib/python3.6/site-packages/azure/functions_worker/dispatcher.py", line 241, in _handle__function_load_request
    function_id, func, func_request.metadata)
  File "/usr/local/lib/python3.6/site-packages/azure/functions_worker/functions.py", line 194, in add_function
    param_bind_type, param_py_type)
  File "/usr/local/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 36, in check_input_type_annotation
    binding = get_binding(binding)
  File "/usr/local/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 20, in get_binding
    registry = get_binding_registry()
  File "/usr/local/lib/python3.6/site-packages/azure/functions_worker/bindings/meta.py", line 13, in get_binding_registry
    return func.get_binding_registry()

I tried rebuilding everything from scratch. Deleted my .env, etc. to no avail. Deleted the app service and recreated it. Nothing helps.

The timers work fine in Visual Studio Code running on my local machine.
Python 3.6.8
CLI 2.0.67
azure-functions==1.0.0a5
azure-functions-worker==1.0.0a6

I dont know what else to do. I would appreciate how to solve it.
Greg

maiqbal11:
We've made changes to remove our dependencies from the `azure` namespace so that we no longer run into conflicts with user-included dependencies. The `azure-functions` package should be kept as part of the `requirements.txt` as it is a dependency needed for user code. We've made changes to our tooling to include `azure-functions` in new apps by default. `azure-functions-worker` should not be included in the `requirements.txt` as it is an implementation detail on our end. Thanks everyone for your patience with this as we iron out the experience while moving from preview to general availability!

twentytwokhz:
Just wanted to point out this also happened to me when targeting an older azure-function version.
The current ones that are listed in the release history are:

_Sept 11, 2019_
`azure-functions==1.0.4`

_Sept 5, 2019_
`azure-functions-worker==1.0.0`

Issue #496, opened 2019-07-23
Azure Functions Python Library 07/23/2019 [breaking] [Consumption Only]


maiqbal11:
With the latest release of the Azure Functions Python Runtime, users who have `azure-functions-worker` included in their `requirements.txt` will need to remove it to avoid breaking their app. Once removed, users will need to republish their application. 

If the dependency is not included, no action is needed. 

Please use Azure/app-service-announcements-discussions#104 for discussion.

PR #506, opened 2019-07-29
Remove azure namespace

maiqbal11:
Removing the worker from the azure namespace. 

Avoids us running into the following sets of issues which are due conflicts with the `azure` namespace in the user's packages:

https://github.com/Azure/azure-functions-python-worker/issues/504
https://github.com/Azure/app-service-announcements-discussions/issues/104


pragnagopa:
Also have to make CLI and docker changes... please link PRs so that we can track changes across components

Hazhzeng:
> Also have to make CLI and docker changes... please link PRs so that we can track changes across components

I don't really think we need to change the dockerfiles and CLI since we change the "worker.py" to use the new **azure_functions_worker** module. The worker.py is the entrypoint, and in the next CLI release, the build should pick up this change and refer to the new module automatically.

pragnagopa:
@maiqbal11 - can you confirm if you are just changing the namespace and not the package name?

maiqbal11:
@pragnagopa, @Hazhzeng I've pushed two additional commits that might be worth taking a look. Basically, they are regarding the decision we had made to cache the current version of `azure.functions` for customers while giving precedence to the user package if they include `azure.functions`. These commits should enable that scenario.

azure_functions_worker/dispatcher.py
```diff
@@ -381,6 +382,7 @@ def _serialize_exception(self, exc):
 
             # Reload azure namespace for customer's libraries
             importlib.reload(sys.modules['azure'])
+            importlib.reload(sys.modules['azure.functions'])
```
pragnagopa: did you test this e2e? i.e - write a python function that imports a different version of azure.funcitons package

Also, add logging 
-  'Reloading azure` also, include version of the package before and after reload
- 'Reloading azure-functions`- same here include version of the package before and after reload
here that has info the version that is loaded.

maiqbal11: Made changes to add logging for when we reload both modules. `azure` itself isn't guaranteed to a have a version or a well defined path, but I've added specific path information for when we reload `azure.functions` so that we can inspect if the path has change from the system to the user path. There is also handling if we are unable to reload `azure.functions` (for when the user hasn't included it). This ensures that we are able to fall back to the default version. 

I was able to validate these changes end to end for the following app scenarios: 

1. No dependencies
2. `azure` dependency 
3. `azure.functions` included in user code and additional `azure` dependencies to see if we are able to resolve properly
4. `azure` native dependency

azure_functions_worker/dispatcher.py
```diff
@@ -376,7 +382,23 @@ def _serialize_exception(self, exc):
                 os.environ[var] = env_vars[var]
 
             # Reload azure namespace for customer's libraries
-            importlib.reload(sys.modules['azure'])
+            logger.info('Reloading azure module')
+            try:
+                importlib.reload(sys.modules['azure'])
+            except Exception as ex:
+                logger.info('Unable to reload azure: \n{}'.format(ex))
+            logger.info('Reloaded azure module')
```
pragnagopa: Move `logger.info('Reloaded azure module')` inside try

pragnagopa:
@maiqbal11 - When is the reload called if app is using placeholders? Did you test with and without placeholders?

maiqbal11:
@pragnagopa, I was able to test with and without placeholders with expected behavior in both cases. The without placeholders case is what we currently have in App Service. For App Service, users will need to include the latest `azure.functions` in their requirements.txt in order to use it. There will not be a concept of having it baked it since we are not loading it during specialization from placeholders. This is the world we eventually want to be in (i.e. users include `azure.functions` themselves) so it might be best to leave it as is).  The alternatives we can consider are:

1. Start in placeholder mode in App Service. Requires changes to the host, but can be done. It is redundant considering App Service has minimal advantages from placeholders. 
2. Move the reload code out of the environment reload request and into some other code path. Specifically, the only other code path where this could possibly live is in the `FunctionLoadRequest` since that this is the only other place where we can guarantee that the user's dependencies have already been downloaded and are on the import path. I'm very hesitant about this because it means that we are doing away with what is a very clean abstraction in favor of getting things to work. Also we would be doing work that we later need to undo when users fully control the `azure-functions` package.

The intention of this work item for me was to be able to publish consistent guidance on what users should do going forward with the `azure.functions` dependency. For both users of App Service and Consumption, we can simply say that they should include `azure-functions` (always pulls latest) or `azure-functions==1.0.0b5` in their requirements.txt. This avoids the confusing messaging that we have for this release (https://github.com/Azure/azure-functions-python-worker/issues/496) which doesn't apply to App Service (see discussion here: https://github.com/Azure/azure-functions-python-worker/issues/504). This would likely be enabled through core tools as well - we would include `azure-functions` by default in the requirements.txt going forward. The default behavior of having a version baked in is just a small benefit we are able to provide for people who are already publishing apps without it. My impression is that users on App Service are publishing with `azure.functions` already since doing anything else would lead to conflicts with our `azure` namespace if they have `azure` in their requirements (also pointed out in https://github.com/Azure/azure-functions-python-worker/issues/504). 

My vote would be to keep things as is and avoid doing further worth that we would later have to undo. With this PR we are able to avoid the largest aspects of the break, but we should still be messaging that people include `azure-functions` in their requirements.txt rather than providing a stop-gap. Sorry for the long-winded response. Lemme know what you think!

pragnagopa:
thanks! yes I agree with the approach we took. No changes needed. Just making sure we tested placholder path as we do not have automated tests.

Issue #411, opened 2019-05-08
How to receive events batch in Eventhub triggered function


joal42:
#### Investigative information
I have a eventhub triggered function in python and want to receive events in batch.

In function.json set `cardinality` to `many`
  "cardinality": "many",
What should the parameter in main function to pass a list of events?

This example can only get one event data:
`def main(event: func.EventHubEvent):`

pragnagopa:
@asavaritayal  - This needs to be documented. Receiving a batch of eventhub events is the recommended approach.
@elprans  - Can you share sample code for receiving an array or list of eventHubEvent?

asavaritayal:
@1st1 @elprans once you provide a sample, I can go update the documentation.

maiqbal11:
While working on a different EventHub issues, came across this issue: https://github.com/Azure/azure-webjobs-sdk/issues/1830. Looks like setting `cardinality` to `many` has the potential to error out if valid json is not provided. Wanted to make a note of this because I ran into this with our default EventHub template in core tools. If we are going to be recommending `cardinality : many` as a way to get batched events, we should at least ensure that the default experience works.

joal42:
@maiqbal11 Thanks for linking the webjobs issue. I set to `many` and use the input as bytes. And then, I convert bytes to array of string. It looks working fine. It would be better if you can share some default usage example.

joal42:
@maiqbal11 I found that I loose eventhub meta data, like `offset` if I use batch model. Could you also help with how to get eventhub meta date in batch processing?

pragnagopa:
For batch processing, trigger data and metadata is received as an array via RPC. Please see java sample code here: https://github.com/Azure/azure-functions-java-worker/blob/dev/endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/EventHubTriggerTests.java#L27
Note: SystemPropertiesArray is populated by the java worker.

@elprans / @maiqbal11  - Does python worker expose an array of metadata?

joal42:
@pragnagopa Thanks for your information. I checked the similar test code path https://github.com/Azure/azure-functions-python-worker/blob/dev/tests/test_eventhub_functions.py
I didn't find the similar metadata array populated.
@elprans @maiqbal11  It would be nice if you can help with it.

maiqbal11:
The SystemPropertiesArray isn't currently being populated by the Python worker. Will need to make a change that allows it to handle this case.

PR #527, opened 2019-08-21
Handle eventhub cardinality

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/411

### DIFF
--- a/tests/endtoend/eventhub_batch_functions/eventhub_multiple/function.json
+++ b/tests/endtoend/eventhub_batch_functions/eventhub_multiple/function.json
@@ -4,17 +4,19 @@
   "bindings": [
     {
       "type": "eventHubTrigger",
-      "name": "event",
+      "name": "events",
       "direction": "in",
-      "eventHubName": "python-worker-eventhub-ci-linux",
+      "cardinality": "many",
+      "dataType": "string",
+      "eventHubName": "python-worker-ci-eventhub-batch",
       "connection": "AzureWebJobsEventHubConnectionString"
     },
     {
-      "type": "blob",
       "direction": "out",
+      "type": "table",
       "name": "$return",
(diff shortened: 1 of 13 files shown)


pragnagopa:
Can you please send a separate PR for updating proto file?

maiqbal11:
@pragnagopa, rebased on top of the new `dev` after updating protobuf.

tests/endtoend/eventhub_batch_functions/eventhub_multiple/function.json
```diff
@@ -0,0 +1,22 @@
+{
+  "scriptFile": "__init__.py",
+
+  "bindings": [
+    {
+      "type": "eventHubTrigger",
+      "name": "events",
+      "direction": "in",
+      "cardinality": "many",
+      "dataType": "string",
```
pragnagopa: do not specify dataType. String is default

maiqbal11:
> no changes needed in the worker or library to convert TypedDataCollection to a list of events? Can you please point me code where this conversion happens?

Here is the corresponding library code: https://github.com/Azure/azure-functions-python-library/pull/24

tests/endtoend/eventhub_batch_functions/eventhub_multiple/function.json
maiqbal11: Looks like we are failing on a JSON deserialization step when `dataType` is not specified as `string`. The SDK must be assuming that we are writing a JSON deserializable string by default. I was able to repro in Node. Will open an issue against the SDK.

maiqbal11: Opened an issue on the EventHub extension: https://github.com/Azure/azure-functions-eventhubs-extension/issues/17.

PR #531, opened 2019-08-26
Changing Fox Colors 🎨🦊

alexkarcher-msft:
The mighty fox is now perfectly color coordinated with the Functions logo.


alexkarcher-msft:
The original readme actually had no logo present, but I can add a Functions logo to the repo and put that in. 

We were hoping that by adding an illustration of a Functions logo, a python logo, and a fun mascot, the repo becomes both more official and more fun.

pragnagopa:
Thanks. Official functions logo similar to https://github.com/Azure/azure-functions-host/ is my preference. Not sure what would be the best place for mascot.

alexkarcher-msft:
Also, quick heads up that the build overall appears broken and blocking me from merging this change into dev.

Everything should be all up to speed on the readme though 🙂

Issue #529, opened 2019-08-23
Module cannot be imported when running remotely but works locally.


altmnt:
I have a function app that uses google storage to pull some information for processing. 
The only dependent package is google.cloud.storage. 
Function works when running locally but fails when running in azure with message
`
Exception: ModuleNotFoundError: No module named 'google.cloud'
`
--build-native-deps, --build remote make no difference. 
pip freeze does not help either.
I downloaded squashfs image - it contains  .python_packages/lib/python3.6/site-packages/google/cloud/storage folder.

#### Investigative information

- Timestamp: Executing 'Functions.svc_machine_learning' (Reason='EventGrid trigger fired at 2019-08-23T07:40:53.1887094+00:00', Id=bdcdce12-be80-4ffd-8434-020823c7393a)
- Function App name: svc_machine_learning
- Core Tools version: 2.7.1575

<details>
<summary>Source</summary>

```python
# __init__.py
import json
import logging

import azure.functions as func


def main(event: func.EventGridEvent):
    result = json.dumps({
        'id': event.id,
        'data': event.get_json(),
        'topic': event.topic,
        'subject': event.subject,
        'event_type': event.event_type,
    })

    logging.info('Python EventGrid trigger processed an event: %s', result)
    from google.cloud import storage
    logging.info('GCS imported!')
    print('GCS imported!')
```
```txt
#requirements.txt
azure-functions
google-cloud-storage
```

</details>

maiqbal11:
@altmnt, I was able to be repro. Looks to be an issue on our end. Will investigate and update.

nkma1989:
Can confirm, I have the same issue. Both "--build remote" and "--build-native-deps" produces this issue.
My requirements.txt are:
azure-functions
cachetools==3.1.1
certifi==2019.6.16
chardet==3.0.4
google-api-core==1.14.2
google-auth==1.6.3
google-cloud-core==1.0.3
google-cloud-logging==1.12.1
googleapis-common-protos==1.6.0
grpcio==1.23.0
idna==2.8
protobuf==3.9.1
pyasn1==0.4.6
pyasn1-modules==0.2.6
pytz==2019.2
requests==2.22.0
rsa==4.0
six==1.12.0
urllib3==1.25.3

I import the following libraries:
import google.cloud.logging
from google.cloud.logging import DESCENDING

nkma1989:
@maiqbal11 I can confirm that adding the following to my code did indeed work.
```import importlib
import sys
[...]

PR #539, opened 2019-09-03
Reload additional modules that user might need

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/529


maiqbal11:
@Hazhzeng, could you take look when you have some time? Should be a straightforward fix.

Issue #519, opened 2019-08-14
Duplicate binding data in gRPC messages


Hazhzeng:
### Scenario
When customer sends a BlobTrigger package with 70MB, the function host will eat up to 210MB.
This is a known issue in function host. Fixes may require in Python worker side.

### Related Issues
https://github.com/Azure/azure-functions-host/issues/4735 (Host)
https://github.com/Azure/azure-functions-host/issues/3845 (Node worker)
https://github.com/Azure/azure-functions-host/issues/3722 (Node worker)

### Action Item on Python Worker
Similar to https://github.com/Azure/azure-functions-nodejs-worker/pull/238/files , we need to check duplicated trigger data. Only populate trigger metadata if it is not populated.

Hazhzeng:
The issue is caused by duplicated data in
1. Request's $req->value
2. TriggerMetadata: req->value
3. TriggerMetadata: $req->value

There are two parts we need to implement:

1. The python worker needs to ask the host to trim off the redundant data in TriggerMetadata
Host Reference: https://github.com/Azure/azure-functions-host/commit/f6aeef80d966a5a6b332328fce38d96c3d7fe582
Node Worker Reference: https://github.com/Azure/azure-functions-nodejs-worker/commit/3b51b6f169fd7a392dd15339d26e5deee60ccbac

2. The python worker needs to reconstruct the TriggerMetadata from request
Host Reference: https://github.com/Azure/azure-functions-host/pull/4660/files
Node Worker Reference: https://github.com/Azure/azure-functions-nodejs-worker/commit/f4b85e4bee

anirudhgarg:
Waiting for merge.

PR #552, opened 2019-09-22
Pass bytes from invocation request metadata

Hazhzeng:
Fix a bug https://github.com/Azure/azure-functions-python-worker/issues/519
1. Forbid host from sending duplicated body bytes data in [input_data, http.rawBody, trigger metadata].
2. Reconstruct http body from input_data.
3. Conversion by content type on [host](https://github.com/Azure/azure-functions-host/blob/96fe7f8fae29f859e4a7d3b40cc6d9c9c3f25d81/src/WebJobs.Script/OutOfProc/Rpc/MessageExtensions/RpcMessageConversionExtensions.cs#L235)

Discussed with @yojagad @pragnagopa

### DIFF
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -217,6 +217,8 @@ class Dispatcher(metaclass=DispatcherMeta):
         capabilities = dict()
         capabilities[constants.RAW_HTTP_BODY_BYTES] = "true"
         capabilities[constants.TYPED_DATA_COLLECTION] = "true"
+        capabilities[constants.RPC_HTTP_BODY_ONLY] = "true"
+        capabilities[constants.RPC_HTTP_TRIGGER_METADATA_REMOVED] = "true"
 
         return protos.StreamingMessage(
             request_id=self.request_id,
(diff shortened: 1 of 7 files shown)


maiqbal11:
@Hazhzeng, can you pull in the host version with these capability changes via the `WEBHOST_URL` in `setup.py`? This will help us get the CI run with the host changes as well.

azure_functions_worker/dispatcher.py
```diff
@@ -217,6 +217,8 @@ def _serialize_exception(self, exc):
         capabilities = dict()
         capabilities[constants.RAW_HTTP_BODY_BYTES] = "true"
```
maiqbal11: You should be able to remove the `RAW_HTTP_BODY_BYTES` capability if you are using the new capability `RPC_HTTP_BODY_ONLY` I believe. They effectively do the same thing from my understanding except that `rawBody` is not longer needed. \cc @yojagad to confirm if this is the case.

Hazhzeng:
Dude, the are those connection strings shared across pipelines? Seems like the UUID from **Azure Functions Python Worker CI** can be read by 
**Azure.e2e-nightly-azure-functions-python-worker**

Issue #450, opened 2019-06-04
Nuget artifact generated by the Python worker is version specific


ankitkumarr:
`azure-functions-python-worker` creates nuget artifact using Python 3.6. This means that all the native binaries are created only for Python 3.6. 

This causes `azure-functions-core-tools` to be unable to support other versions of Python for Functions in Python. This will be true for all apps, because `grpc` has native binaries, for example - `cygrpc.cpython-36m-darwin.so` in `grpc._cython` that is used by the worker.

As you can see from the example above, because Python version 36 is specified above (because the binary was created for 36), this gets rejected by any other Python interpreter (say 3.7) giving this error - 
`ImportError: cannot import name 'cygrpc' from 'grpc._cython'`

@asavaritayal, @maiqbal11, @anirudhgarg I am not sure what the best solution might be, but this will be a blocker from allowing any users to publish apps from multiple Python version.

Maybe we should go back to installing `azure.functions_worker` on user's machine directly.

Update: It's not just Python version, they would also need to have the target architecture (64 bit) as well.

maiqbal11:
@pragnagopa, @fabiocav as well, this might tie into the earlier discussion we had about supporting multiple versions of dependencies for different platforms via the `worker.config.json`. Do we also want include a versioning for languages within that? Something like `<language>.<os>.<architecture_version>.<language_version>`? We can then choose which version of the worker to run based on the user specifying the worker runtime language **and** its version?

pragnagopa:
We need to start creating issues for adding support for 3.7 addressing this issue will be part of updating tooling to support 3.7

ankitkumarr:
Had a chat with @pragnagopa, ideally we would want to be thinking about solving the bigger problem of supporting multiple versions everywhere (local, azure, etc). This should be kept in mind before we make any decisions.

Until we have resolved that problem, we could have this workaround --

Publish only 3.6 compatible nuget with the host / core-tools. If users have 3.7 or other versions that are not supported, link them to the docs about how to manually install 3.7 version worker and replace the existing one locally. We would point them to the path `workers/python/deps`, ask them to clear it out and have them `pip install --target` it there. That would also mean for core-tools to be able to know what version is the worker on the path compatible for (3.6 or 3.7). 

Instead of doing it manually, it can be automated via core-tools but we don't want to implement this workaround or a hack if the core issue will be addressed later.

anirudhgarg:
[...]

PR #555, opened 2019-09-27
Generate version/os/arch bits in NuGet

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/450. Will also need to make host changes in https://github.com/Azure/azure-functions-host/issues/4935


ankitkumarr:
@maiqbal11, do you think we'd be able to modify the target such that for each platform/architecture, only those python worrker binaries will be copied?

This will greatly help reduce the size of both the Host and the core-tools.

maiqbal11:
> @maiqbal11, do you think we'd be able to modify the target such that for each platform/architecture, only those python worrker binaries will be copied?
> 
> This will greatly help reduce the size of both the Host and the core-tools.

@ankitkumarr, could you elaborate more on where you'd expect this modified targets file to be applicable? Right now, judging by the Host artifacts that we generate, we don't have ones that have per-platform/architecture bits. We can take that change in a separate PR if needed. I don't want to block this one since it's been open for a while and we have other changes that need this.

ankitkumarr:
@maiqbal11, I was thinking about maybe changing the target a little bit here--
https://github.com/Azure/azure-functions-python-worker/pull/555/files#diff-289dcafb488dcabd264ce602bd8ecaccR17-R18

Here, we might be able to use either -- 
1. `$(OS)` ([reference](https://stackoverflow.com/questions/43412140/csproj-copy-files-depending-on-operating-system)) to only copy from `$(OS)` subdir, or 
2. Have separate targets for each arch/platform with a `Condition`like --
`<Copy ... Condition=" '$(OS)' == 'Unix'>`

I haven't tried any of the above, so I can't confidently say that it would work. But reading the docs, it should.
I am ok with it being a separate item. It may make a difference to the size of our Docker images given that it may include all the arch and os with the current design (if size is an issue).

maiqbal11:
> @maiqbal11, I was thinking about maybe changing the target a little bit here--
> https://github.com/Azure/azure-functions-python-worker/pull/555/files#diff-289dcafb488dcabd264ce602bd8ecaccR17-R18
> 
> Here, we might be able to use either --
> 
> 1. `$(OS)` ([reference](https://stackoverflow.com/questions/43412140/csproj-copy-files-depending-on-operating-system)) to only copy from `$(OS)` subdir, or
> 2. Have separate targets for each arch/platform with a `Condition`like --
>    `<Copy ... Condition=" '$(OS)' == 'Unix'>`
> 
> I haven't tried any of the above, so I can't confidently say that it would work. But reading the docs, it should.
> I am ok with it being a separate item. It may make a difference to the size of our Docker images given that it may include all the arch and os with the current design (if size is an issue).

@ankitkumarr, I see your concern and it's definitely something that I agree with. This would require some exploration to see how it meshes with the current build process for generating Host bits. I'm somewhat familiar with what the Host CI does but would need to see how the per-platform story plays out. Just given the current timelines, I would prefer to address that as a separate work item.

python/prod/worker.py
```diff
@@ -0,0 +1,66 @@
...
+import platform
+import subprocess
+from pathlib import Path
+
+# User packages
+PKGS_PATH = "site/wwwroot/.python_packages"
+VENV_PKGS_PATH = "site/wwwroot/worker_venv"
+
+PKGS_36 = "lib/python3.6/site-packages"
+PKGS = "lib/site-packages"
```
ankitkumarr: assuming `/` instead of `os.path.sep` or `os.path.join` is ok as this will only be used in Azure.

ankitkumarr:
@maiqbal11, I think that sounds fair. Thanks for making this change.

pack/Microsoft.Azure.Functions.PythonWorker.nuspec
```diff
@@ -0,0 +1,24 @@
...
+    <copyright>© .NET Foundation. All rights reserved.</copyright>
+  </metadata>
+  <files>
+    <file src="..\3.6_WINDOWS_x64\**" target="tools\3.6\WINDOWS\X64" />
+    <!-- <file src="..\3.6_WINDOWS_X86\**" target="tools\3.6\WINDOWS\X86" /> -->
+    <file src="..\3.6_LINUX_X64\**" target="tools\3.6\LINUX\X64" />
+    <file src="..\3.6_OSX_X64\**" target="tools\3.6\OSX\X64" />
+    <file src="..\3.7_WINDOWS_X64\**" target="tools\3.7\WINDOWS\X64" />
+    <file src="..\3.7_WINDOWS_X86\**" target="tools\3.7\WINDOWS\X86" />
+    <file src="..\3.7_LINUX_X64\**" target="tools\3.7\LINUX\X64" />
```
ankitkumarr: @maiqbal11, looks like https://github.com/Azure/azure-functions-host/pull/4985/files#diff-41f6f4a743bf773c5edaa3da72f2df34R28 is not all upper case. It may cause issues as Linux file system is case sensitive.

cc: @Hazhzeng

maiqbal11: The `OSPlatform.ToString()` method is documented to be upper case: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.osplatform.tostring?view=netcore-3.0. We should be fine here.

python/prod/worker.py
maiqbal11: Yes, that's right. I wanted to define these as constants to make it clear that this is only for the Linux environment in Azure.

pack/Microsoft.Azure.Functions.PythonWorker.nuspec
ankitkumarr: Oh perfect! Thanks for clarifying.

Issue #547, opened 2019-09-17
Move user packages' initialization to worker 


maiqbal11:
With https://github.com/Azure/azure-functions-host/issues/4935, the Python workers for different versions/platforms/architectures will be packaged with the host. We should no longer need the logic in [python-context](https://github.com/Azure/azure-functions-docker/blob/master/host/2.0/stretch/amd64/python-context) to initialize the Python worker environment and run the worker. The logic in [start.sh](https://github.com/Azure/azure-functions-docker/blob/master/host/2.0/stretch/amd64/python-context/start.sh) should move to the worker to ensure that user packages are still registered for their app version (3.6 or 3.7).

maiqbal11:
Fixed via https://github.com/Azure/azure-functions-python-worker/pull/555

PR #556, opened 2019-09-27
Add start up code to worker

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/547

### DIFF
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -1,3 +1,10 @@
 # Capabilities
 RAW_HTTP_BODY_BYTES = "RawHttpBodyBytes"
 TYPED_DATA_COLLECTION = "TypedDataCollection"
+
+# User packages
+PKGS_PATH = "site/wwwroot/.python_packages"
+VENV_PKGS_PATH = "site/wwwroot/worker_venv"
+
+PKGS_36 = "lib/python3.6/site-packages"
+PKGS_37 = "lib/python3.7/site-packages"
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -2,9 +2,13 @@
 RAW_HTTP_BODY_BYTES = "RawHttpBodyBytes"
 TYPED_DATA_COLLECTION = "TypedDataCollection"
 
+# Azure environment variables
+AZURE_WEBSITE_INSTANCE_ID = "WEBSITE_INSTANCE_ID"
+AZURE_CONTAINER_NAME = "CONTAINER_NAME"
+
 # User packages
 PKGS_PATH = "site/wwwroot/.python_packages"
 VENV_PKGS_PATH = "site/wwwroot/worker_venv"
 
 PKGS_36 = "lib/python3.6/site-packages"
-PKGS_37 = "lib/python3.7/site-packages"
+PKGS = "lib/site-packages"
--- a/azure_functions_worker/main.py
+++ b/azure_functions_worker/main.py
@@ -30,6 +30,11 @@ def parse_args():
     return parser.parse_args()
 
 
+def is_azure_environment():
+    return (constants.AZURE_CONTAINER_NAME in os.environ or
+            constants.AZURE_WEBSITE_INSTANCE_ID in os.environ)
+
+
 def setup_user_pkg_paths():
     minor_version = sys.version_info[1]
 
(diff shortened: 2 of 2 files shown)


azure_functions_worker/constants.py
```diff
@@ -1,3 +1,10 @@
 # Capabilities
 RAW_HTTP_BODY_BYTES = "RawHttpBodyBytes"
 TYPED_DATA_COLLECTION = "TypedDataCollection"
+
+# User packages
+PKGS_PATH = "site/wwwroot/.python_packages"
+VENV_PKGS_PATH = "site/wwwroot/worker_venv"
+
+PKGS_36 = "lib/python3.6/site-packages"
+PKGS_37 = "lib/python3.7/site-packages"
```
Hazhzeng: In PGKS_PATH, we want to drop the version specific package path.
We can add a python path to support **lib/site-packages** as well, so that we will have a transition windows for updating our toolings.

azure_functions_worker/main.py
```diff
@@ -26,8 +30,28 @@ def parse_args():
...
+    home = Path.home()
+    pkgs_path = os.path.join(home, constants.PKGS_PATH)
+    venv_pkgs_path = os.path.join(home, constants.VENV_PKGS_PATH)
+
+    if minor_version == 6:
+        sys.path.insert(1, os.path.join(venv_pkgs_path, constants.PKGS_36))
+        sys.path.insert(1, os.path.join(pkgs_path, constants.PKGS_36))
+    elif minor_version == 7:
+        sys.path.insert(1, os.path.join(venv_pkgs_path, constants.PKGS_37))
+        sys.path.insert(1, os.path.join(pkgs_path, constants.PKGS_37))
```
ankitkumarr: out of curiosity, why 1 and not 0 for these?

ankitkumarr:
With this change, we would update the path even if we are running the worker locally. Is that expected? If there's a way to tell whether or not we are running it in Azure context, maybe we can use that for updating the path.

Probably not a big deal though, as invalid entries in path are usually harmless.

Hazhzeng:
Thanks @maiqbal11 for reminding me of this.

Python 3.6 Paths:
- site/wwwroot/.python_packages/lib/python3.6/site-packages
- site/wwwroot/.python_packages/lib/site-packages
- site/wwwroot/worker_venv/lib/python3.6/site-packages
- site/wwwroot/worker_venv/lib/site-packages

Python 3.7 Paths:
- site/wwwroot/.python_packages/lib/python3.7/site-packages
- site/wwwroot/.python_packages/lib/site-packages
- site/wwwroot/worker_venv/lib/python3.7/site-packages
- site/wwwroot/worker_venv/lib/site-packages

azure_functions_worker/main.py
Hazhzeng: We may want the currently directory path to have the top priority.

maiqbal11: I had wanted to keep the current directory path (i.e. `''`) as the first search path. This mimics the current behavior of the `PYTHONPATH` which is appended as the second entry in the `sys.path` after the current directory.

ankitkumarr:
@Hazhzeng, thanks!

I think we only need --

Python 3.6 Paths:
- site/wwwroot/.python_packages/lib/python3.6/site-packages
- site/wwwroot/.python_packages/lib/site-packages
- site/wwwroot/worker_venv/lib/python3.6/site-packages

Python 3.7 Paths:
- site/wwwroot/.python_packages/lib/site-packages

Hoping that no point of time would we need to deploy a virtual environment.
Is there a reason you included the others?

Hazhzeng:
> @Hazhzeng, thanks!
> 
> I think we only need --
> 
> Python 3.6 Paths:
> 
> * site/wwwroot/.python_packages/lib/python3.6/site-packages
> * site/wwwroot/.python_packages/lib/site-packages
> * site/wwwroot/worker_venv/lib/python3.6/site-packages
> 
> Python 3.7 Paths:
> 
> * site/wwwroot/.python_packages/lib/site-packages
> 
> Adding, `site/wwwroot/.python_packages/lib/python3.7/site-packages` as a safety net for the future is fine. But not sure if we need others.
> Hoping that no point of time would we need to deploy a virtual environment.
> Is there a reason you included the others?

One of my concern for **worker_venv** is in the Azure DevOps Python deployment page https://docs.microsoft.com/en-us/azure/azure-functions/functions-how-to-azure-devops#python.

Currently, we ask our customers to resolve the dependencies by running **python3.6 -m venv worker_venv**.

Desperately, the virtual environment path is different depends on the OS on the devops agent. 
On windows: worker_venv\Lib\site-packages
On ubuntu: worker_venv/lib/python3.6/site-packages

In conclusion, we only need to support `site/wwwroot/.python_packages/lib/site-packages` path for Python 3.7 and future versions.

maiqbal11:
@ankitkumarr, @Hazhzeng, please take look. I believe all comments should be addressed.

maiqbal11:
Closing this in favor of consolidating all changes under https://github.com/Azure/azure-functions-python-worker/pull/555. It makes sense of the generated NuGet to contain these worker changes since they are essential for the new NuGet to work. I've created a separate commit for the changes in this PR here: https://github.com/Azure/azure-functions-python-worker/pull/555/commits/107d6952bca9ea1ef88201e99f781124ad87c20a. I needed to included all the startup code in `worker.py` because adding to `sys.path` directly was not registering user packages. We need to add them to the `PYTHONPATH` before we spin up a new python process with the worker.

PR #570, opened 2019-11-01
Separate run paths for Azure and local development

maiqbal11:


python/prod/worker.py
```diff
@@ -42,25 +42,21 @@ def determine_user_pkg_paths():
...
-    if platform.system() == 'Windows':
-        joined_pkg_paths = ";".join(user_pkg_paths)
-        env['PYTHONPATH'] = f'{joined_pkg_paths};{func_worker_dir}'
-        # execve doesn't work in Windows: https://bugs.python.org/issue19124
-        subprocess.run([sys.executable,
-                       '-m', 'azure_functions_worker'] + sys.argv[1:],
-                       env=env)
-    else:
         joined_pkg_paths = ":".join(user_pkg_paths)
         env['PYTHONPATH'] = f'{joined_pkg_paths}:{func_worker_dir}'
```
Hazhzeng: I think we should use `os.pathsep` instead of `":"`

pragnagopa:
Please create a release with these changes and also add also, release notes

maiqbal11:
> Please create a release with these changes and also add also, release notes

Done: https://github.com/Azure/azure-functions-python-worker/releases/tag/1.0.1

PR #581, opened 2019-11-21
Update library to 1.0.5 for CI

maiqbal11:


Hazhzeng:
@maiqbal, do we need to bump up the worker version?

maiqbal11:
> @maiqbal, do we need to bump up the worker version?

We'll need to do a worker release to generate a new NuGet that contains the latest library code. This is because we only generate the NuGet (used in core tools plus the Host) from the worker CI.

PR #587, opened 2019-12-02
Add python 3.8 nuget generation

Hazhzeng:
Add python 3.8 into matrix for build and test
- Bump flake to 3.7.9
- Fix linting for code for flake 3.7.9
- Remove setup_requires (because we need the grpcio to be installed. We don't need its .egg packages, and it will mess up Python 3.8 build process if we keep the setup_requires).
- Add Python 3.8 worker in build matrix

Miscellaneous:
- Add .env*/ and .venv*/ in .gitignore
- Ignore .env*/ and .venv*/ in flake8


Hazhzeng:
@maiqbal11 I've separated the package generation for V2 and V3. Please take a look.
https://azfunc.visualstudio.com/Azure%20Functions%20Python/_build/results?buildId=3472&view=artifacts&type=publishedArtifacts

pack/Microsoft.Azure.Functions.V2.PythonWorker.nuspec
```diff
@@ -18,7 +18,11 @@
     <file src="..\3.7_WINDOWS_X86\**" target="tools\3.7\WINDOWS\X86" />
     <file src="..\3.7_LINUX_X64\**" target="tools\3.7\LINUX\X64" />
     <file src="..\3.7_OSX_X64\**" target="tools\3.7\OSX\X64" />
+    <!-- <file src="..\3.8_WINDOWS_X64\**" target="tools\3.8\WINDOWS\X64" /> -->
```
maiqbal11: nit: We can just remove these. Is there a reason why you wanted to leave these as comments?

Hazhzeng: Yep, I can remove it, just for the ease of comparison with v3 python worker during development.

PR #554, opened 2019-09-26
Accept TraceContext from host

yojagad:
Part of fix for Azure/azure-functions-host#3747
Host PR: Azure/azure-functions-host#4917
LanguageWorker PR: Azure/azure-functions-language-worker-protobuf#40

Older PR which was closed: https://github.com/Azure/azure-functions-python-worker/pull/545 in favor of this one

PR #601, opened 2020-01-07
Update proto and add custom build steps

maiqbal11:
Needed to unblock https://github.com/Azure/azure-functions-python-worker/pull/554. Contains the same logic as https://github.com/Azure/azure-functions-python-worker/pull/528 which was reverted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,8 @@ import sys
 import tempfile
 import urllib.request
 import zipfile
+import re
+from distutils import dir_util
 from distutils.command import build
 
 from setuptools import setup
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -73,56 +75,74 @@ def _gen_grpc(self):
...
-        built_proto_files_dir = (built_protos_dir
-                                 / 'azure_functions_worker' / 'protos')
+        built_protos_dir = build_dir / 'built_protos'
 
-        if os.path.exists(build_dir):
+        if os.path.exists(staging_root_dir):
             shutil.rmtree(build_dir)
 
         if os.path.exists(built_protos_dir):
             shutil.rmtree(built_protos_dir)
```
Hazhzeng: Seems like these four lines can be combined together?
```python
if os.path.exists(build_dir):
    shutil.rmtree(build_dir)
```
Am I missing something here?

setup.py
```diff
@@ -73,56 +75,74 @@ def _gen_grpc(self):
...
                   file=sys.stderr)
             sys.exit(1)
 
-        for f in compiled:
-            shutil.copy(f, proto_root_dir)
+        # Needed to support absolute imports in files. See
+        # https://github.com/protocolbuffers/protobuf/issues/1491
+        self.make_absolute_imports(compiled_files)
+
+        dir_util.copy_tree(built_protos_dir, str(proto_root_dir))
```
Hazhzeng: Do we still need **built_protos_dir**? I think we should clean it up after we build it.

setup.py
```diff
@@ -73,56 +75,74 @@ def _gen_grpc(self):
...
-        if os.path.exists(build_dir):
+        if os.path.exists(staging_root_dir):
             shutil.rmtree(build_dir)
 
         if os.path.exists(built_protos_dir):
             shutil.rmtree(built_protos_dir)
 
-        proto_files = glob.glob(str(proto_src_dir / '**' / '*.proto'),
-                                recursive=True)
+        shutil.copytree(proto_src_dir, staging_dir)
```
Hazhzeng: Doesn't seems like **staging_dir** is used anywhere else. Maybe we can remove it?

maiqbal11: It's being used to do the copying as you pointed here. In the actual build command, we are using the same directory structure without explicitly referencing the `staging_dir`. Might be useful to walk through this offline as it's slightly tedious to explain here.

PR #603, opened 2020-01-09
Enable Python 3.8 in V3 worker

Hazhzeng:
1. Fix the nuget.targets name to reflect the nuget package.
2. Separate V2/V3 Python worker.config.json 
  - for V2 worker, we only enable 3.6 and 3.7
  - for V3 worker, we enable 3.6, 3.7 and 3.8


python/prodV3/worker.py
```diff
@@ -0,0 +1,61 @@
...
+    pkgs_path = os.path.join(home, PKGS_PATH)
+    venv_pkgs_path = os.path.join(home, VENV_PKGS_PATH)
+
+    user_pkg_paths = []
+    if minor_version == 6:
+        user_pkg_paths.append(os.path.join(venv_pkgs_path, PKGS_36))
+        user_pkg_paths.append(os.path.join(pkgs_path, PKGS_36))
+        user_pkg_paths.append(os.path.join(pkgs_path, PKGS))
+    elif minor_version == 7:
+        user_pkg_paths.append(os.path.join(pkgs_path, PKGS))
```
maiqbal11: Should also include a check here for 3.8: `if minor_version == 7 or minor_version == 8`

maiqbal11:
@Hazhzeng, looks like this was related to the names not matching up then? Building locally with this configuration lead to properly binplacing in V3?

Hazhzeng:
@maiqbal11 
Yeah, I renamed the .targets file yesterday in nuget cache and completely rebuild the project, the bin placing works fine then.

PR #606, opened 2020-01-10
Update proto packages with identity and shared

maiqbal11:
Added packages that will contain the newly introduced `identity` and `shared` folders so they are correctly packaged when publishing.


Hazhzeng:
This PR is cherry picked in https://github.com/Azure/azure-functions-python-worker/pull/605

PR #613, opened 2020-01-29
Change current working directory when reloading environment variables

Hazhzeng:
This can solve the issue when user tries to import files in the function app directory.
They should no longer require to use **os.path.abspath** for deriving the file paths.

Added rollback flag **PYTHON_ROLLBACK_CWD_PATH** for disabling the cwd change behavior


Hazhzeng:
As discussed with @AnatoliB,
1. We need to add a rollback flag for this feature in case it breaks our customer
2. We need to announce the change we make in this change

azure_functions_worker/dispatcher.py
```diff
@@ -426,6 +431,14 @@ def _serialize_exception(self, exc):
                 request_id=self.request_id,
                 function_environment_reload_response=failure_response)
 
+    @disable_feature_by(constants.PYTHON_ROLLBACK_CWD_PATH)
```
Hazhzeng: This function will not executed when PYTHON_ROLLBACK_CWD_PATH is set to `true` or `1`.

Hazhzeng:
Hey @AnatoliB, this PR is ready for review. Feel free to come to me for additional questions. Thanks in advance.

azure_functions_worker/utils/common.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def is_true_like(setting: str):
+    if setting is None:
+        return False
+
+    return setting.lower().strip() in ['1', 'true', 't', 'yes', 'y']
+
+
+def is_false_like(setting: str):
```
AnatoliB: You don't seem to need `is_false_like` and `is_envvar_false` yet, other than for symmetry. Consider removing unnecessary code for now.

azure_functions_worker/dispatcher.py
```diff
@@ -408,6 +409,11 @@ def _serialize_exception(self, exc):
                 logger.info('Unable to reload azure.functions. '
                             'Using default. Exception:\n{}'.format(ex))
 
+            # Change function app directory
+            if getattr(func_env_reload_request, 'function_app_directory'):
```
AnatoliB: Isn't `getattr(func_env_reload_request, 'function_app_directory')` equivalent to `func_env_reload_request.function_app_directory` (including the fact that both will raise the same `AttributeError` if `function_app_directory` is not provided)? If the intent here to gracefully handle the absence of `function_app_directory`, you probably need to provide a default value. Otherwise, why not just use the simpler code: `func_env_reload_request.function_app_directory`?

Hazhzeng: Yea, good catch, I will add a default value **None** when function_app_directory is absence.

Issue #608, opened 2020-01-16
Error due to output binding checks in dispatcher.py


priyaananthasankar:
The Durable Orchestration trigger typically has a function.json that looks like this

```
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "name": "context",
      "type": "orchestrationTrigger",
      "direction": "in",
      "dataType": "string"
    }
  ],
  "disabled": false
}
```
This does not have output binding and so it fails in dispatcher.py with this error message:
 **f'function {fi.name!r} without a $return binding '**

Added these checks but these are causing regression in HTTP Triggers and other further scenarios for Durable
```
#if call_result is not None and not fi.has_return:
            #    raise RuntimeError(
            #        f'function {fi.name!r} without a $return binding '
            #        f'returned a non-None value')
 
```
And we have added an additional check in lines 344-349 where if the return type is None, we just define a dummy return binding and send it back
 
```
else:
                ret_binding_name = "ret_binding_name"
                ret_pytype = typing.Any
                return_value = bindings.to_outgoing_proto(
                    ret_binding_name, call_result,
                    pytype=ret_pytype)
 ```

Need a fix in dispatcher.py that does not give the error with the given function.json above.

How to reproduce: Use the function.json above

Issue #227, opened 2018-10-20
Durable Functions Support for Python


asavaritayal:
New Feature - Looking for votes and/or user input to gauge traction.

priyaananthasankar:
+1 Absolutely needed. Working on a small sample with asyncio to simulate something similar.

yokawasa:
+1 Yes needed!  Python has many choice of data science libraries and it definitely has long running & stateful scenario!

tjhgit:
I am experimenting with triggering an azure batch job from an azure function. So this is completely asynchronous and you can also scale up the compute ressources easily in azure batch.

mataralhawiti:
+1 Absolutely

SaschaSchlemmer:
+1 for long running analytics/ML jobs

Sarah-Aly:
This is absolutely a needed feature. I do have a customer who uses Python in Azure Functions for manipulating CSVs & creating time series with Pandas and they are an ideal case for function chaining in DF.

t-eckert:
+1 Happy to contribute

nroypf:
+1 Need it asap :)

dajor:
+1

FlippieCoetser:
currently running large data transformations with Python. Execution time is around 30 min. Having durable functions will help a lot!

ericdrobinson:
I'm saddened to see that Python support for Durable Functions isn't available yet. I've chronicled my efforts to make my usage of the HTTPTrigger more effective/efficient for my workflow in the comments of issue #236. The documentation, however, [keeps suggesting](https://docs.microsoft.com/en-us/azure/azure-functions/functions-best-practices#avoid-long-running-functions) that I take the [Durable Functions route](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-concepts#async-http) for a more reasonable experience.

Unfortunately, it looks as though I'll need to re-implement Durable Functions manually if I want to gain the suggested benefits of that approach :(

+1, indeed.

explora21:
+1, of course

SpicySyntax:
+1, This would be very useful for long running ML Workloads

reynoldsa:
Absolutely +1.

bigdatamoore:
+1 agreed.  We could use this now.

jaryder:
+1 - Is there a timeline for when this will be available?

svartkanin:
+1 - Definitely needed

[...]

PR #623, opened 2020-02-21
Add durable functions support

Hazhzeng:
### Background
We want to enable durable functions in Azure Functions Python Worker.
Previously, we are blocked by the implicit return from the generic binding.
The solution in this PR is to define a rich binding for OrchestrationTrigger and allow implicit returns from the trigger.

### Depends On
https://github.com/Azure/azure-functions-python-library/pull/48
CI is failing due to dependencies on azure-functions-python-library

### Resolves
resolves: https://github.com/Azure/azure-functions-python-worker/issues/608
resolves: https://github.com/Azure/azure-functions-python-worker/issues/227

### Usage
```javascript
// function.json
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "name": "orctx",
      "type": "orchestrationTrigger",
      "direction": "in"
    }
  ],
  "disabled": false
}
```

```python
# main.py
def main(orctx: func.OrchestrationContext):
    orchestrate = df.Orchestrator.create(generator_function)
    result = orchestrate(orctx.body)
    return result
```

### DIFF
--- a/azure_functions_worker/bindings/__init__.py
+++ b/azure_functions_worker/bindings/__init__.py
@@ -2,6 +2,7 @@ from .tracecontext import TraceContext
 from .context import Context
 from .meta import check_input_type_annotation
 from .meta import check_output_type_annotation
+from .meta import has_implicit_output
 from .meta import is_trigger_binding
 from .meta import from_incoming_proto, to_outgoing_proto
 from .out import Out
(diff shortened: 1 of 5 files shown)


Hazhzeng:
@scgbear Yes, this is due to dependency on azure-functions-python-library

asedighi:
i verified the code with @Hazhzeng and tested the faninfanout sample with durable + retry.  I confirmed that it worked.

PR #624, opened 2020-02-28
Add unittests for generic bindings and durable functions

Hazhzeng:
Waiting for the durable functions library to be released for the full E2E test
1. Testing if the generic binding is working properly when
     a. When function.json **dataType** is set to 'string' or 'bytes'
     b. When function.json **type** has a generic binding name (does not match any of our built-in triggers)
     c. When main.py has proper or no annotation type
2. Ensure durable function's **activityTrigger** and **orchestrationTrigger** does not fall into any of the generic bindings


ankitkumarr:
25 checks?  :O

PR #649, opened 2020-04-15
[Build] Fix unittest cases on Windows and MacOS

Hazhzeng:


tests/unittests/test_loader.py
```diff
@@ -80,7 +80,11 @@ async def _runner():
 
             stdout, stderr = await proc.communicate()
 
-            self.assertEqual(stdout.strip().split(b'\n'), [b'True', b'True'])
+            # Trimming off carriage return charater when testing on Windows
+            stdout_lines = [
+                l.replace(b'\r', b'') for l in stdout.strip().split(b'\n')
+            ]
```
vrdmr, line 86: Good catch.

setup.py
```diff
@@ -32,23 +31,23 @@
   <ItemGroup>
     <PackageReference
         Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator"
-        Version="1.0.1"
+        Version="1.1.7"
```
vrdmr, line 34: Thanks for doing this! :)

Issue #714, opened 2020-07-04
[BUG] Python Function app down with Microsoft.Azure.WebJobs.Script.Workers.WorkerProcessExitException


TsuyoshiUshio:
I raise this issue for sharing workaround. 

###### Please provide the following:
Service is down . When I try to browse function, the portal loads indefinitely without being able to load: 
Then show `Azure Functions Runtime is unreachable`. 
Server logs showing: 

```
Stopping JobHost
Change analysis service stopped before analysis completion. Waiting for cancellation
RuntimeError: no currently running Dispatcher is found
raise RuntimeError('no currently running Dispatcher is found')
File "/azure-functions-host/workers/python/3.6/LINUX/X64/azure_functions_worker/dispatcher.py", line 39, in current
Dispatcher.current._on_logging(record, msg)
File "/azure-functions-host/workers/python/3.6/LINUX/X64/azure_functions_worker/dispatcher.py", line 525, in emit
```

The other log I found
```
Microsoft.Azure.WebJobs.Script.Workers.WorkerProcessExitException Exception Count: 2 Source:Worker.rpcWorkerProcess.python.b63de34f-2111-4b23-9829-87e523c6a7de Stacktrace: 
python exited with code 1 self._log(ERROR, msg, args, **kwargs), raise RuntimeError('no currently running Dispatcher is found'),RuntimeError: 
```

This issue is my customer's. I can't share their FunctionAppName.
- Timestamp: 2020-07-02 16:49 
- Function Runtime Version: 2
- Python

###### Provide the steps required to reproduce the problem:
N/A
###### Provide a description of the expected behavior.

Function host start successfully.

###### Provide a description of the actual behavior observed.

Start to fail as I showed. 

###### Provide a description of any known workarounds.

Change the `linux-fx-version` to the old version. 

```
az webapp config set -g YOUR_RESOURCE_GROUP -n YOUR_FUNCTION_APP_NAME --linux-fx-version "DOCKER|mcr.microsoft.com/azure-functions/python:2.0.13351-python3.6-appservice"
```

PR #715, opened 2020-07-06
Fix no currently running Dispatcher is found issue

Hazhzeng:
## Description

Fixes https://github.com/Azure/azure-functions-python-worker/issues/714
- Ensure the messages in the gRPC async logging handler are flushed when encountering exceptions
- Ensure the messages in the sys.stdout are flushed before disabling console logging
- Redirect exception in AsyncLoggingHandler emit method into host console, since
    - When encounter dispatcher exception, the logging will be switched to console log
    - If not suppressing the exception in AsyncLoggingHandler emit method, it will trigger an infinite exception logging in gRPC channel.
- Never suppress error console logging handler, since it redirects log to stderr. This should be always enabled.

Tests:
1. Add a new test case for hijacking current event loop, since I see a lot of Azure Packages (`azure-eventhubs`, `azure-identity` and `azure-keyvault-secrets`) are using `asyncio.get_event_loop()` to acquire current event loop and hijacking the context. We should be aware of that and check if our worker is compatible with these kind of operations.
2. Add a new test case for `print()` logging from customer's code.

Miscellaneous:
1. Change _UnixSelectorEventLoop to BaseEventLoop for annotation since _UnixSelectorEventLoop is private and it is not supported in Windows.
2. Run sys.stdout.flush() after each customer function call, since now we use gRPC to log entries generated from customer's logger. We should maintain the behavior of `print()` logging.
3. Simplify the logic in `is_system_log_category` when testing if a log entry belongs to system.log category
---

### PR information

- [X] The title of the PR is clear and informative.
- [X] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [X] If applicable, the PR references the bug/issue that it fixes in the description.
- [X] New Unit tests were added for the changes made and CI is passing.

- [X] I have read the [contribution guidelines](https://github.com/Azure/azure-functions-python-worker/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -23,6 +23,7 @@ from . import loader
 from . import protos
 from . import constants
 
+from .constants import CONSOLE_LOG_PREFIX
 from .logging import error_logger, logger, is_system_log_category
 from .logging import enable_console_logging, disable_console_logging
 from .utils.tracing import marshall_exception_trace
@@ -129,13 +130,13 @@ class Dispatcher(metaclass=DispatcherMeta):
             try:
                 await forever
             finally:
-                logger.info('Detaching gRPC logging due to exception.')
+                logger.warn('Detaching gRPC logging due to exception.')
                 logging_handler.flush()
                 root_logger.removeHandler(logging_handler)
 
                 # Reenable console logging when there's an exception
                 enable_console_logging()
-                logger.info('Switched to console logging due to exception.')
+                logger.warn('Switched to console logging due to exception.')
         finally:
             DispatcherMeta.__current_dispatcher__ = None
 
@@ -170,7 +171,7 @@ class Dispatcher(metaclass=DispatcherMeta):
 
         if is_system_log_category(record.name):
             log_category = protos.RpcLog.RpcLogCategory.System
-        else:  # import logging; logging.info will yield 'root' in record.name
+        else:  # customers using logging will yield 'root' in record.name
             log_category = protos.RpcLog.RpcLogCategory.User
 
         log = dict(
@@ -538,11 +539,13 @@ class AsyncLoggingHandler(logging.Handler):
         msg = self.format(record)
         try:
             Dispatcher.current.on_logging(record, msg)
-        except RuntimeError:
+        except RuntimeError as runtime_error:
             # This will cause 'Dispatcher not found' failure.
             # Logging such of an issue will cause infinite loop of gRPC logging
-            # To mitigate, we should suppress the 2nd level error logging here.
-            pass
+            # To mitigate, we should suppress the 2nd level error logging here
+            # and use print function to report exception instead.
+            print(f'{CONSOLE_LOG_PREFIX} ERROR: {str(runtime_error)}',
+                  file=sys.stderr, flush=True)
 
 
 class ContextEnabledTask(asyncio.Task):
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -27,7 +27,7 @@ from .logging import error_logger, logger, is_system_log_category
 from .logging import enable_console_logging, disable_console_logging
 from .utils.tracing import marshall_exception_trace
 from .utils.wrappers import disable_feature_by
-from asyncio.unix_events import _UnixSelectorEventLoop
+from asyncio import BaseEventLoop
 from logging import LogRecord
 from typing import Optional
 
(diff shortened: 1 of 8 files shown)


azure_functions_worker/dispatcher.py
```diff
@@ -48,7 +48,7 @@ class Dispatcher(metaclass=DispatcherMeta):
 
     _GRPC_STOP_RESPONSE = object()
 
-    def __init__(self, loop: _UnixSelectorEventLoop, host: str, port: int,
+    def __init__(self, loop: BaseEventLoop, host: str, port: int,
```
vrdmr, line 51: Thanks for fixing this.

azure_functions_worker/dispatcher.py
```diff
@@ -113,22 +113,29 @@ async def dispatch_forever(self):
...
             finally:
-                # Re-enable console logging when there's an exception
-                enable_console_logging()
+                logger.info('Detaching gRPC logging due to exception.')
+                logging_handler.flush()
                 root_logger.removeHandler(logging_handler)
+
+                # Reenable console logging when there's an exception
+                enable_console_logging()
+                logger.info('Switched to console logging due to exception.')
```
vrdmr, line 138: Can this be a warning - we need to make it clear that console logging is not the standard anymore.

azure_functions_worker/dispatcher.py
```diff
@@ -113,22 +113,29 @@ async def dispatch_forever(self):
...
-            logger.info('Detach console logging. Switch to gRPC logging')
+            logger.info('Switched to gRPC logging.')
+            logging_handler.flush()
 
             try:
                 await forever
             finally:
-                # Re-enable console logging when there's an exception
-                enable_console_logging()
+                logger.info('Detaching gRPC logging due to exception.')
```
vrdmr, line 132: Same as below - warning.

azure_functions_worker/dispatcher.py
```diff
@@ -163,7 +170,7 @@ def on_logging(self, record: logging.LogRecord, formatted_msg: str) -> None:
 
         if is_system_log_category(record.name):
             log_category = protos.RpcLog.RpcLogCategory.System
-        else:
+        else:  # import logging; logging.info will yield 'root' in record.name
```
vrdmr, line 173: Can you clarify why is this comment added?

azure_functions_worker/dispatcher.py
```diff
@@ -518,10 +528,21 @@ def gen(resp_queue):
...
+        # buffered in this handler, not calling the emit yet.
         msg = self.format(record)
-        Dispatcher.current.on_logging(record, msg)
+        try:
+            Dispatcher.current.on_logging(record, msg)
+        except RuntimeError:
+            # This will cause 'Dispatcher not found' failure.
+            # Logging such of an issue will cause infinite loop of gRPC logging
+            # To mitigate, we should suppress the 2nd level error logging here.
+            pass
```
vrdmr, line 545: Can we add a telemetry event - something which we can see in Kusto, without logging? For my info.

azure_functions_worker/logging.py
```diff
@@ -51,25 +59,19 @@ def setup(log_level, log_destination):
...
 def disable_console_logging() -> None:
+    # We should only remove the sys.stdout stream, as error_logger is used for
+    # unexpected critical error logs handling.
     if logger and handler:
+        handler.flush()
         logger.removeHandler(handler)
 
-    if error_logger and error_handler:
-        error_logger.removeHandler(error_handler)
-
```
vrdmr, line 59: So we are removing the two different handlers - `handler` and `error_handler`?

azure_functions_worker/dispatcher.py
Hazhzeng: A lot of users using our basic template
```python
import logging
logging.info('some message')
```
Will use the **root** logger for logging, so we mark `record.name == 'root'` as protos.RpcLog.RpcLogCategory.User. This comment is here to explain why root logger belongs to User logs.

azure_functions_worker/logging.py
Hazhzeng: Previously, we removed error_handler. But when I tested today, I think we should preserve error_logger, since sys.stderr will always picked up by host. This PR here is to stop removing error_handler.

PR #744, opened 2020-09-23
Add PYTHON_THREADPOOL_THREAD_COUNT app setting

### LINKED ISSUE #675: Investigate making the number of threads in the Azure Functions Python Worker thread pool configurable
Currrently the Azure Functions Python Worker restricts the number of threads in the default thread pool to 1. We want to experiment with possibly offering a configurable values for the customer to change this to a higher value.

There is already a detailed discussion here: https://github.com/Azure/azure-functions-python-worker/issues/236 if it makes sense to allow multiple threads in the Python Worker thread pool. 

This is the line of code which has to be changed.        

https://github.com/Azure/azure-functions-python-worker/blob/6068092982479932f1d2d3c315ee6835cdeac8c1/azure/functions_worker/dispatcher.py#L56

Hazhzeng:
## Description

This is a dependency of enabling async thread pool. Getting this change into the codebase will allow us configure the threadpool more easily when benchmarking the async threadpool performance.

Fixes #675

---

Now we allow customer to configure number of thread in the synchronous threadpool by setting PYTHON_THREADPOOL_THREAD_COUNT application setting. If the application settings is not set, we will default it to 1.

I'm wondering whether we should call this PYTHON_SYNC_THREADPOOL_THREAD_COUNT. 
If we introduce async thread pool later on, should we introduce a new PYTHON_ASYNC_THREADPOOL_THREAD_COUNT app settings? It is possible that if a function app mixes both sync and async functions, the actual number of threads will be 2x. Adding this new setting will has a better the transparency for customer to control their function app, but will make the app setting more complicate.

In my opinion, both async threadpool and sync threadpool should use the value in PYTHON_THREADPOOL_THREAD_COUNT.

### PR information

- [X] The title of the PR is clear and informative.
- [X] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [X] If applicable, the PR references the bug/issue that it fixes in the description.
- [X] New Unit tests were added for the changes made and CI is passing.

- [X] I have read the [contribution guidelines](https://github.com/Azure/azure-functions-python-worker/blob/master/.github/CONTRIBUTING.md).


azure_functions_worker/dispatcher.py
```diff
@@ -62,17 +63,11 @@ def __init__(self, loop: BaseEventLoop, host: str, port: int,
...
-        # async support for that.)
+        # We allow the customer to change synchronous thread pool count by
+        # PYTHON_THREADPOOL_THREAD_COUNT app setting. The default value is 1.
+        thread_count = get_app_setting(PYTHON_THREADPOOL_THREAD_COUNT, '1')
         self._sync_call_tp = concurrent.futures.ThreadPoolExecutor(
-            max_workers=1)
+            max_workers=int(thread_count))
 
         self._grpc_connect_timeout = grpc_connect_timeout
         # This is set to -1 by default to remove the limitation on msg size
```
vrdmr, line 73: For the below variables without type_hinting, can you please add them as well?

```python
self._grpc_connect_timeout: float = grpc_connect_timeout
# This is set to -1 by default to remove the limitation on msg size

self._grpc_max_msg_len: int = grpc_max_msg_len
self._grpc_thread: Optional[threading.Thread] = threading.Thread(name='grpc-thread', target=self.__poll_grpc)
```

vrdmr: Also please add a log-line in the Connect

```
logger.info('Max Workers %s', disp._sync_call_tp._max_workers)
```

vrdmr: And changing `warn` to `warning`

azure_functions_worker/dispatcher.py
```diff
@@ -62,17 +63,11 @@ def __init__(self, loop: BaseEventLoop, host: str, port: int,
...
-        # only run one synchronous function in parallel.  This is
-        # because synchronous code in Python is rarely designed with
-        # concurrency in mind, so we don't want to allow users to
-        # have races in their synchronous functions.  Moreover,
-        # because of the GIL in CPython, it rarely makes sense to
-        # use threads (unless the code is IO bound, but we have
-        # async support for that.)
+        # We allow the customer to change synchronous thread pool count by
+        # PYTHON_THREADPOOL_THREAD_COUNT app setting. The default value is 1.
+        thread_count = get_app_setting(PYTHON_THREADPOOL_THREAD_COUNT, '1')
```
vrdmr, line 68: Currently, this is only sync threads - make this explicit in the variable name.

```suggestion
        sync_tp_max_workers_count = get_app_setting(PYTHON_THREADPOOL_THREAD_COUNT, '1')
```

vrdmr: Also, make it `int` here itself

azure_functions_worker/constants.py
```diff
@@ -15,6 +15,7 @@
 
 # Feature Flags (app settings)
 PYTHON_ROLLBACK_CWD_PATH = "PYTHON_ROLLBACK_CWD_PATH"
+PYTHON_THREADPOOL_THREAD_COUNT = "PYTHON_THREADPOOL_THREAD_COUNT"
```
vrdmr, line 18: @stefanushinardi - Could you please check if you want to change the name?

azure_functions_worker/dispatcher.py
```diff
@@ -62,24 +63,20 @@ def __init__(self, loop: BaseEventLoop, host: str, port: int,
...
-        # use threads (unless the code is IO bound, but we have
-        # async support for that.)
-        self._sync_call_tp = concurrent.futures.ThreadPoolExecutor(
-            max_workers=1)
-
-        self._grpc_connect_timeout = grpc_connect_timeout
+        # We allow the customer to change synchronous thread pool count by
+        # PYTHON_THREADPOOL_THREAD_COUNT app setting. The default value is 1.
+        self._sync_tp_max_workers: int = int(get_app_setting(
+            PYTHON_THREADPOOL_THREAD_COUNT, '1'))
```
vrdmr, line 69: If the customer puts in an incorrect value, it might throw an exception when casting to int - you will need to take care of that - try-except, etc.

azure_functions_worker/dispatcher.py
```diff
@@ -90,6 +87,8 @@ async def connect(cls, host: str, port: int, worker_id: str,
         disp._grpc_thread.start()
         await disp._grpc_connected_fut
         logger.info('Successfully opened gRPC channel to %s:%s', host, port)
+        logger.info('Python worker sync threadpool max workers: %s',
```
vrdmr, line 90: If possible, can we merge the two log statements into one? :) Trying to reduce logging that happens from the Python worker side.

tests/unittests/test_dispatcher.py
```diff
@@ -0,0 +1,95 @@
...
+            self.assertIsNotNone(invoke_id)
+            self.assertEqual(call_r.response.result.status,
+                             protos.StatusResult.Success)
+
+            # Ensure the dispatcher sync threadpool count is set to 1
+            self.assertEqual(ctrl._worker._sync_tp_max_workers, 5)
+
+    async def test_dispatcher_sync_threadpool_invalid_worker(self):
+        '''Test when sync threadpool maximum worker is set to an invalid value,
+        the host should fail to start
```
vrdmr, line 86: Do we want to fail or do we want to default to 1? I prefer not failing but making it clear to the customer when running functions that there is an invalid value for the Application Settings.

@anirudhgarg - Any thoughts?

vrdmr: This is related to my previous comment - 

> If the customer puts in an incorrect value, it might throw an exception when casting to int - you will need to take care of that - try-except, etc.

azure_functions_worker/dispatcher.py
vrdmr: Also - I feel that we should keep a true max - like 32 or 64? We don't want customers to put in 10000 in there (I understand it is thus a theoretical value and python only creates a new thread when it needs it and not immediately)

tests/unittests/test_dispatcher.py
```diff
@@ -0,0 +1,95 @@
...
+                    )
+                ])
+            self.assertIsNotNone(invoke_id)
+            self.assertEqual(call_r.response.result.status,
+                             protos.StatusResult.Success)
+
+            # Ensure the dispatcher sync threadpool count is set to 1
+            self.assertEqual(ctrl._worker._sync_tp_max_workers, 1)
+
+    async def test_dispatcher_sync_threadpool_set_worker(self):
```
vrdmr, line 51: You might also want to add a `check_logs_*` test as well to see if the logline came in from the host.

vrdmr (changes requested):

azure_functions_worker/dispatcher.py
Hazhzeng: I will cap the max value to 32, check the 3.8 implementation below https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

tests/unittests/test_dispatcher.py
Hazhzeng: OK, I will add a validator field in get_app_setting

tests/unittests/test_dispatcher.py
Hazhzeng: The testutils.AsyncTestCase case does not have check_logs_ utils in it. I will create another test class to validate the error logs.

Hazhzeng: The check_logs_* check does not work here since the threadpool initialization starts before the function invocation.
The check_log host_out only records the invocation context because the `self.host_stdout.seek(last_pos)` the cursor keeps updating from the last test. https://github.com/Azure/azure-functions-python-worker/blob/dev/azure_functions_worker/testutils.py#L225.

Thus, I think the best I can do is to test the logging with a mock logger.

azure_functions_worker/dispatcher.py
```diff
@@ -475,7 +473,28 @@ def _change_cwd(self, new_cwd: str):
...
+    def _get_sync_tp_max_workers(self):
+        def tp_max_workers_validator(value: str):
+            try:
+                int_value = int(value)
+            except ValueError:
+                logger.warning(f'{PYTHON_THREADPOOL_THREAD_COUNT} must be an '
+                               'integer')
+                return False
+
+            if not 1 <= int_value <= 32:
```
vrdmr, line 487: nit; make 32 as a constant

azure_functions_worker/dispatcher.py
```diff
@@ -475,7 +473,28 @@ def _change_cwd(self, new_cwd: str):
             os.chdir(new_cwd)
             logger.info('Changing current working directory to %s', new_cwd)
         else:
-            logger.warn('Directory %s is not found when reloading', new_cwd)
+            logger.warning('Directory %s is not found when reloading', new_cwd)
+
+    def _get_sync_tp_max_workers(self):
```
vrdmr, line 478: nit; could you please add return datatype? for the inner function as well?

azure_functions_worker/dispatcher.py
```diff
@@ -475,7 +473,28 @@ def _change_cwd(self, new_cwd: str):
...
+            if not 1 <= int_value <= 32:
+                logger.warning(f'{PYTHON_THREADPOOL_THREAD_COUNT} must be set '
+                               'to a value between 1 and 32')
+                return False
+
+            return True
+
+        return int(get_app_setting(
+            setting=PYTHON_THREADPOOL_THREAD_COUNT,
+            default_value='1',
```
vrdmr, line 496: nit; actually - could you also have this as a constant at the top - making 1 explicit as a default when reading.

azure_functions_worker/utils/common.py
```diff
@@ -15,3 +16,46 @@ def is_envvar_true(env_key: str) -> bool:
         return False
 
     return is_true_like(os.environ[env_key])
+
+
+def get_app_setting(
+    setting: str,
+    default_value: Optional[str] = None,
+    validator: Optional[Callable[[str], bool]] = None
```
vrdmr, line 24: Better approach! :)

tests/unittests/test_dispatcher.py
vrdmr: ah - sounds good. Thanks for the info.

vrdmr (approved):
LGTM.

some nits.

azure_functions_worker/constants.py
stefanushinardi: LGTM - putting my reasoning here for future reference: it makes sense to put the PYTHON_* prefix for this as without it users might get confused and use this config for other languages

vrdmr (approved):

PR #749, opened 2020-10-02
Reduce logging for functions invocations

vrdmr:
## Description
Reduce logging for functions invocations to optimize for performance. This reduces the # of log statements from 3 to 1 (merging the first 2 log lines and removing the successful invocation)

---

### PR information

### DIFF
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -7,37 +7,35 @@ Implements loading and execution of Python workers.
 
 import asyncio
 import concurrent.futures
-import logging
-import queue
-import threading
-import os
-import sys
 import importlib
 import inspect
+import logging
+import os
+import queue
+import sys
+import threading
+from asyncio import BaseEventLoop
+from logging import LogRecord
+from typing import Optional
 
 import grpc
 
 from . import bindings
+from . import constants
 from . import functions
 from . import loader
 from . import protos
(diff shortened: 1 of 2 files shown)


azure_functions_worker/dispatcher.py
```diff
@@ -7,37 +7,35 @@
...
+from .logging import disable_console_logging, enable_console_logging
+from .logging import error_logger, is_system_log_category, logger
 from .utils.common import get_app_setting
 from .utils.tracing import marshall_exception_trace
 from .utils.wrappers import disable_feature_by
-from asyncio import BaseEventLoop
-from logging import LogRecord
-from typing import Optional
+
+_TRUE = "true"
```
Hazhzeng, line 38: Put it in constants.py would be a better place.

Hazhzeng (approved):
Thanks for cleaning this up, neat!

azure_functions_worker/dispatcher.py
vrdmr: Constants.py seem to be specific to more config-level or global setting and keeping constants within the modules they are used is a better approach.

If the constant is to be used in multiple modules, constants.py is better and we can refactor this when we see it being used in other places.

Hazhzeng: Ah, I see, thanks for the explanation!

PR #765, opened 2020-10-14
Shared memory data transfer between Functions Host and Python worker

gohar94:
## Description

Fixes azure/azure-functions-host#6791

---

### PR information


gohar94:
cc: @goiri 

Opening a draft PR for now. You can skim through whenever.

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +107,54 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
+                param_binding = protos.ParameterBinding(
+                                    name=out_name,
+                                    rpc_shared_memory=shmem)
+            else:
+                raise Exception(
+                    'cannot write datum value into shared memory'
+                )
+        else:
+            raise Exception(
+                'unsupported datum type for shared memory'
```
goiri, line 151: Add the type and probably the name to the exception.

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +107,54 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
+
+
+def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
+                      pytype: typing.Optional[type],
+                      out_name: str,
+                      shmem_mgr: SharedMemoryManager,
+                      invocation_id: str) -> protos.ParameterBinding:
+    datum = get_datum(binding, obj, pytype)
+    # TODO gochaudh: IMPORTANT: Right now we set the AppSetting to disable this
+    # However that takes impact only for data coming from host -> worker
```
goiri, line 128: Would it be possible to reply with shared memory only if the input was shared memory?

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +107,54 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
+                    name=map_name,
+                    offset=0,
+                    count=len(value),
+                    type=protos.RpcSharedMemoryDataType.bytes)
+                param_binding = protos.ParameterBinding(
+                                    name=out_name,
+                                    rpc_shared_memory=shmem)
+            else:
+                raise Exception(
+                    'cannot write datum value into shared memory'
```
goiri, line 147: Add name of the binding to the error.

azure_functions_worker/bindings/meta.py
```diff
@@ -82,8 +95,8 @@ def from_incoming_proto(
             f'and expected binding type {binding}')
 
 
-def to_outgoing_proto(binding: str, obj: typing.Any, *,
-                      pytype: typing.Optional[type]) -> protos.TypedData:
+def get_datum(binding: str, obj: typing.Any,
+              pytype: typing.Optional[type]):
```
goiri, line 99: return type?

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +107,54 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
+                      shmem_mgr: SharedMemoryManager,
+                      invocation_id: str) -> protos.ParameterBinding:
+    datum = get_datum(binding, obj, pytype)
+    # TODO gochaudh: IMPORTANT: Right now we set the AppSetting to disable this
+    # However that takes impact only for data coming from host -> worker
+    # Is there a way to check the AppSetting here so that this does not respond back
+    # with shared memory?
+    param_binding = None
+    if shmem_mgr.is_enabled() and shmem_mgr.is_supported(datum):
+        if datum.type == 'bytes':
```
goiri, line 133: Don't we have the option for string?
We could potentially also support the current stream too.

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,74 @@
...
+from ..logging import logger
+from ..mmap_handler.file_writer import FileWriter
+from ..mmap_handler.file_reader import FileReader
+from ..mmap_handler.file_accessor import FileAccessor
+
+
+class SharedMemoryManager:
+    """
+    Performs all operations related to reading/writing data from/to Shared
+    Memory.
```
goiri, line 14: As this is in the Python worker we could describe a little the exchange of data between C# and the worker.

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def get_bytes(self, mmap_name: str, offset: int, count: int) -> bytes:
+        """
+        Reads data from the given Memory Mapped File with the provided name,
+        starting at the provided offset and reading a total of count bytes.
+        Returns a tuple containing the binary data read from shared memory
+        if successful, None otherwise.
+        """
+        logger.info('Reading from shared memory: %s', mmap_name)
+        data = FileReader.read_content_as_bytes(mmap_name, offset)
```
goiri, line 46: We probably should handle the count too.

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,74 @@
...
+        Writes the given data into shared memory.
+        Returns the name of the Memory Mapped File into which the data was
+        written if succesful, None otherwise.
+        """
+        mmap_name = str(uuid.uuid4())
+        logger.info('Writing to shared memory: %s', mmap_name)
+        mmap = FileWriter.create_with_content_bytes(mmap_name, data)
+
+        if invocation_id not in self.allocated_mmaps:
+            self.allocated_mmaps[invocation_id] = []
```
goiri, line 60: If we eventually make the binding process async or multi threaded could we have some unexpected race?

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,74 @@
...
+        Reads data from the given Memory Mapped File with the provided name,
+        starting at the provided offset and reading a total of count bytes.
+        Returns a tuple containing the binary data read from shared memory
+        if successful, None otherwise.
+        """
+        logger.info('Reading from shared memory: %s', mmap_name)
+        data = FileReader.read_content_as_bytes(mmap_name, offset)
+        return data
+
+    def put_bytes(self, data: bytes, invocation_id: str) -> str:
```
goiri, line 49: Instead of invocation_id should it be something more generic?

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,74 @@
...
+from ..mmap_handler.file_accessor import FileAccessor
+
+
+class SharedMemoryManager:
+    """
+    Performs all operations related to reading/writing data from/to Shared
+    Memory.
+    """
+    def __init__(self):
+        self.allocated_mmaps = {} # type dict[string, [(mmap_name, mmap)]
```
goiri, line 17: We should probably explain what this represents and the protocol followed

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,74 @@
...
+        Returns the name of the Memory Mapped File into which the data was
+        written if succesful, None otherwise.
+        """
+        mmap_name = str(uuid.uuid4())
+        logger.info('Writing to shared memory: %s', mmap_name)
+        mmap = FileWriter.create_with_content_bytes(mmap_name, data)
+
+        if invocation_id not in self.allocated_mmaps:
+            self.allocated_mmaps[invocation_id] = []
+        self.allocated_mmaps[invocation_id].append((mmap_name, mmap))
```
goiri, line 61: We still have open the discussion between invocation_id and mmap_name release.
However, either way we should have two separate maps.
invocation_id -> [mmap_name]
mmap_name -> mmap

azure_functions_worker/dispatcher.py
```diff
@@ -330,10 +333,12 @@ async def _handle__invocation_request(self, req):
...
                 else:
                     trigger_metadata = None
+
                 args[pb.name] = bindings.from_incoming_proto(
-                    pb_type_info.binding_name, pb.data,
+                    pb_type_info.binding_name, pb,
                     trigger_metadata=trigger_metadata,
-                    pytype=pb_type_info.pytype)
+                    pytype=pb_type_info.pytype,
+                    shmem_mgr=self._shmem_mgr)
```
goiri, line 341: This thing of passing the shared memory manager all over (also applies to the C# code) is getting a little repetitive.
Isn't there a better option? Putting it in a context? Making it a singleton?

azure_functions_worker/mmap_handler/file_accessor.py
```diff
@@ -0,0 +1,128 @@
...
+            except Exception as ex:
+                print.error("Cannot create dir '%s': %s" % (dirname, str(ex)))
+
+    @staticmethod
+    def _create_mmap_file_linux(map_name, map_size):
+        """Get the file descriptor for a new memory map.
+        """
+        escaped_map_name = urllib.parse.quote_plus(map_name)
+        dir_exists = False
+        for mmap_temp_dir in consts.TEMP_DIRS:
```
goiri, line 72: Remember to handle running out of space.

azure_functions_worker/mmap_handler/file_accessor.py
```diff
@@ -0,0 +1,128 @@
+# -*- coding: utf-8 -*-
```
goiri, line 1: In the __init__.py of this folder (for some reason I cannot leave comments there), write a description with pointers to the C# code and probably the issue/docu.

azure_functions_worker/bindings/meta.py
gohar94: That would not handle cases where if the function doesn't take any input (say just an HTTP invocation) and produces a large output that needs to go back from worker to host over shared memory. What we need is a way to enable and disable both the worker and host with the same switch (which I think is going to be an environment variable or some App Settings thingy - will check with folks when we release this PR.)

azure_functions_worker/dispatcher.py
gohar94: I don't see use of globals/singletons in this repo so not sure if that's something people would want. But we can get input on that once we put this in review.

azure_functions_worker/bindings/shared_memory_manager.py
gohar94: When we go there then we would have to make all accesses to `allocated_mem_maps` thread-safe by doing locking or similar. For now, keeping it simple.

azure_functions_worker/bindings/datumdef.py
```diff
@@ -92,6 +94,66 @@ def from_typed_data(cls, td: protos.TypedData):
...
+        """
+        mmap_name = shmem.name
+        offset = shmem.offset
+        count = shmem.count
+        data_type = shmem.type
+        if data_type == protos.RpcDataType.bytes:
+            val = shmem_mgr.get_bytes(mmap_name, offset, count)
+            if val is not None:
+                return cls(val, 'bytes')
+        elif data_type == protos.RpcDataType.string:
```
goiri, line 111: Should we handle stream?

azure_functions_worker/bindings/datumdef.py
```diff
@@ -92,6 +94,66 @@ def from_typed_data(cls, td: protos.TypedData):
...
+                shmem = protos.RpcSharedMemory(
+                    name=map_name,
+                    offset=0,
+                    count=len(value),
+                    type=protos.RpcDataType.string)
+                return shmem
+            else:
+                raise Exception(
+                    f'cannot write datum value (type: {datum.type}) into '
+                    f'shared memory (name: {map_name})'
```
goiri, line 151: map_name will be None in this case, right?

azure_functions_worker/bindings/datumdef.py
```diff
@@ -92,6 +94,66 @@ def from_typed_data(cls, td: protos.TypedData):
...
+            value = datum.value
+            map_name = shmem_mgr.put_string(value)
+            if map_name is not None:
+                shmem = protos.RpcSharedMemory(
+                    name=map_name,
+                    offset=0,
+                    count=len(value),
+                    type=protos.RpcDataType.string)
+                return shmem
+            else:
```
goiri, line 148: Avoid the else, as the if already returns.
(This is a typical pylint comment, you may want to run it, it will give a lot of non-sense suggestions but good to check.)

azure_functions_worker/bindings/datumdef.py
```diff
@@ -92,6 +94,66 @@ def from_typed_data(cls, td: protos.TypedData):
...
+                raise Exception(
+                    f'cannot write datum value (type: {datum.type}) into '
+                    f'shared memory (name: {map_name})'
+                )
+        else:
```
goiri, line 153: No need for the else

azure_functions_worker/bindings/datumdef.py
```diff
@@ -113,4 +175,4 @@ def datum_as_proto(datum: Datum) -> protos.TypedData:
     else:
         raise NotImplementedError(
             'unexpected Datum type: {!r}'.format(datum.type)
-        )
+        )
```
goiri, line 178: Avoid.

azure_functions_worker/bindings/meta.py
```diff
@@ -82,8 +92,8 @@ def from_incoming_proto(
             f'and expected binding type {binding}')
 
 
-def to_outgoing_proto(binding: str, obj: typing.Any, *,
-                      pytype: typing.Optional[type]) -> protos.TypedData:
+def get_datum(binding: str, obj: typing.Any,
```
goiri, line 95: We may want to add a docstring to distinguish it from the old to_outgoing_proto

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +104,39 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
             f'unable to encode outgoing TypedData: '
             f'unsupported type "{binding}" for '
             f'Python type "{type(obj).__name__}"')
+    return datum
+
```
goiri, line 108: Is the code style in the rest of the file to add two breaklines?

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +104,39 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
+    # with shared memory?
+    shared_mem_value = None
+    parameter_binding = None
+    # If shared memory is enabled, try to transfer to host over shared memory
+    if shmem_mgr.is_enabled() and shmem_mgr.is_supported(datum):
+        shared_mem_value = datumdef.Datum.to_rpc_shared_memory(datum, shmem_mgr)
+    if shared_mem_value is not None:
+        # Check if data was transferred over shared memory.
+        # If it was, then use the rpc_shared_memory field in the response message. 
+        parameter_binding = protos.ParameterBinding(
```
goiri, line 133: I'm always torn by singleflow vs multiple returns but given that this code already does multiple returns in other places, I think we should return directly and do return None in the default case.

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,117 @@
...
+        self.file_accessor = FileAccessorFactory.create_file_accessor()
+        self.file_reader = FileReader()
+        self.file_writer = FileWriter()
+
+    def is_enabled(self) -> bool:
+        """
+        Whether supported types should be transferred between functions host and the worker using
+        shared memory.
+        """
+        return True
```
goiri, line 38: Do we want to check the app settings or whichever mechanism we follow in the C# side?

azure_functions_worker/bindings/meta.py
```diff
@@ -94,5 +104,39 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
+    datum = get_datum(binding, obj, pytype)
     return datumdef.datum_as_proto(datum)
+
+
+def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
+                      pytype: typing.Optional[type],
+                      out_name: str,
+                      shmem_mgr: SharedMemoryManager) -> protos.ParameterBinding:
+    datum = get_datum(binding, obj, pytype)
+    # TODO gochaudh: IMPORTANT: Right now we set the AppSetting to disable this
```
goiri, line 121: Should we also check the handshake where it advertises the capabilities?

azure_functions_worker/mmap_handler/file_reader.py
```diff
@@ -0,0 +1,95 @@
...
+
+import mmap
+import os
+import struct
+from typing import Optional
+from .memorymappedfile_constants import MemoryMappedFileConstants as consts
+from .file_accessor_factory import FileAccessorFactory
+
+
+class FileReader:
```
goiri, line 12: Given that on Windows the mmap is probably not a file, I'm not sure File is the best naming.

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,117 @@
...
+        memory.
+        """
+        if datum.type == 'bytes':
+            # TODO gochaudh: Check for min size config
+            # Is there a common place to put configs shared b/w host and worker?
+            # Env variable? App Setting?
+            return True
+        elif datum.type == 'string':
+            return True
+        else:
```
goiri, line 52: no else and just return False at the end?

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,117 @@
...
+            return False
```
goiri, line 53: Handle stream?

azure_functions_worker/bindings/shared_memory_manager.py
```diff
@@ -0,0 +1,117 @@
...
+    end-to-end latency.
+    """
+    def __init__(self):
+        # The allocated memory maps are tracked here so that a reference to them is kept open until
+        # they have been used (e.g. if they contain a function's output, it is read by the
+        # functions host).
+        # Having a mapping of the name and the memory map is then later used to close a given
+        # memory map by its name, after it has been used.
+        # key: map_name, val: mmap.mmap
+        self.allocated_mem_maps: Dict[str, mmap.mmap] = {}
```
goiri, line 28: This is a little convoluted, but what if the worker restarts and the mmaps stay there?
Should we try to repopulate them?

azure_functions_worker/mmap_handler/file_accessor.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Optional
+
+
+class FileAccessor(metaclass=abc.ABCMeta):
+    """
+    For accessing memory maps.
+    This is an interface that must be implemented by sub-classes to provide platform-specific
+    support for accessing memory maps.
+    Currently the following two sub-classes are implemented:
+        1) FileAccessorWindows
```
goiri, line 15: Windows is not a file right?

azure_functions_worker/mmap_handler/file_accessor.py
```diff
@@ -0,0 +1,55 @@
...
+        raise NotImplementedError
+
+    def _verify_new_map_created(self, map_name: str, mem_map) -> bool:
+        """Checks if the first byte of the memory map is zero.
+        If it is not, this memory map already existed.
+        """
+        mem_map.seek(0)
+        byte_read = mem_map.read(1)
+        is_new_mmap = False
+        if byte_read != b'\x00':
```
goiri, line 50: Make it a constant? With a comment explaining that memory map with 0 at the beginning doesn't exist.
BTW, we cannot get a healthy map with the first byte being 0 right?

azure_functions_worker/bindings/datumdef.py
gohar94: There's no stream being exposed here - if you look in `datumdef.py`, those are all the types.
Basically the "stream" is already fully read by the functions host before passing here so it's just bytes now.

azure_functions_worker/bindings/meta.py
gohar94: For some reason, in this file it is :/

azure_functions_worker/bindings/shared_memory_manager.py
gohar94: As mentioned earlier, no streams here.

azure_functions_worker/bindings/shared_memory_manager.py
gohar94: Ummm... that's a good point. If the worker restarts in case of error, then there is already trouble. 
The reason why I don't want to repopulate this is because if the memory map was allocated by the functions host or a different worker process, then things can get weird. 
So, we need to think on how to avoid such leaks if the worker restarts.
But then, what if the functions host restarts - even that could possibly leave maps around, right?

azure_functions_worker/mmap_handler/file_accessor.py
gohar94: I think the reasoning here (which you thought of) was that for a new map, the first byte *will* be `\x00`. And if the map is already being used then it will be something other than that (like we have the header bytes which will make it non-zero) - but let me double check this.

azure_functions_worker/bindings/shared_memory_manager.py
gohar94: Yes, I have to find out if they have some API for doing this in the Python side or I should just check the env variable.

azure_functions_worker/mmap_handler/file_accessor.py
gohar94: Yes, will rename to `SharedMemoryMapAccessor*`.
I am also thinking of refactoring this similar to what we have in C# i.e. have a class called `SharedMemoryMap` and do all the operations which are being done in `FileReader`/`FileWriter` as part of that class.

azure_functions_worker/mmap_handler/file_accessor.py
gohar94: image
Yes, so when we create a memory map and put the header on it, the header bytes look like that ^. The first byte is non-zero so this logic seems to be fine.

azure_functions_worker/mmap_handler/file_reader.py
gohar94: Agreed - will change to `SharedMemoryMap`.

azure_functions_worker/bindings/meta.py
gohar94: Not sure I understand - explain?

azure_functions_worker/bindings/meta.py
gohar94: But this isn't going to return None in any case - it's either `ParameterBinding` with `rpc_shared_memory` or `data`.

azure_functions_worker/bindings/datumdef.py
goiri: RpcDataType does have it (there's also one called http which I wasn't aware) is that implicitly mapped to binary in this part of the code?

azure_functions_worker/bindings/datumdef.py
```diff
@@ -92,6 +94,55 @@ def from_typed_data(cls, td: protos.TypedData):
...
+    @classmethod
+    def to_rpc_shared_memory(cls, datum: Datum, shmem_mgr: SharedMemoryManager) -> protos.RpcSharedMemory:
+        """
+        Writes the given value to shared memory and returns the corresponding RpcSharedMemory
+        object which can be sent back to the functions host over RPC.
+        """
+        if datum.type == 'bytes':
+            value = datum.value
+            map_name = shmem_mgr.put_bytes(value)
+            if map_name is not None:
```
goiri, line 126: The else of this if would end up into a not implemented for a datum type while it is not the issue here.

azure_functions_worker/bindings/meta.py
goiri: Right now from C# we check the capabilities of the python worker; which is the case we would expect while rollout.
I know is unlikely but do we handle the host C# side not supporting shared memory but Python yes?

azure_functions_worker/bindings/meta.py
goiri: True.

azure_functions_worker/bindings/shared_memory_manager.py
goiri: Yes, I'm pretty sure it would be fine but worth giving a though to those situations.

azure_functions_worker/mmap_handler/memorymappedfile_constants.py
```diff
@@ -0,0 +1,19 @@
...
+
+    # The length of a long which is the length of the header in the content memory map
+    CONTENT_LENGTH_NUM_BYTES = 8
+    # The length of the header: content length
+    CONTENT_HEADER_TOTAL_BYTES = CONTENT_LENGTH_NUM_BYTES
+
+    # Zero byte.
+    # E.g. Used to compare the first byte of a newly created memory map against this; if it is a
+    # non-zero byte then the memory map was already created.
+    ZERO_BYTE = b'\x00'
```
goiri, line 19: I would call it EMPTY_MAP_MARK and then in the comment we say this.

goiri: And what does the first byte contain in principle that guarantees a non-0 first byte? Is there any corner case like if the content has a specific lenght that happens to generate a first byte as 0?

gohar94:
Closed and moved to [this](https://github.com/Azure/azure-functions-python-worker/pull/816) PR.

Issue #642, opened 2020-04-01
How to enrich the logger with custom properties?


empz:
Running some Azure Function (runtime 3) with Python 3.7 and Application Insights configured.

Doing logging.info("message") works. This get sent as a TRACE to AppInsights. But it's not very useful if we can't enrich an execution's logging context with custom properties so that we can easily query by these afterwards. 

Is this even supported on Azure Functions for Python? 

What's the relation between Python Azure Functions and the [Opencensus library which is shown as the recommended AI SDK for Python](https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python)?

Or should I just use a 3rd party logging framework (i.e. Sentry, Datadog, etc)

tuckeryatesjvion:
Custom log properties are very important to me as well. I'd love to be able to push relevant identifiers to the customDemensions field in applicationinsights traces, then be able to easily isolate logs related to those identifiers. 

I've had some trouble searching for documentation on this feature, although I suspect it is supported. Can any of the MSFT team link anything relevant?

PR #815, opened 2021-02-03
Python Worker Extension Interface (worker)

Hazhzeng:
## Changes
1. Add ExtensionManager in **extension.py** which can be enabled by the feature flag **PYTHON_ENABLE_WORKER_EXTENSIONS**
2. Add **trace_parent**, **trace_state** and **attributes** replicating **TraceParent**, **TraceState** and **Attributes** values. This change will ensure the worker code is more Pythonic.
3. Improve logging in **worker_init_request**, now the log entry will include information like **Python version**, **Worker version** for ease of diagnostics. 
    `Received WorkerInitRequest, python version 3.7.5 (tags/v3.7.5:5c02a39a0b, Oct 15 2019, 00:11:34) [MSC v.1916 64 bit (AMD64)], worker version 1.1.10, request ID 768c03e6-b194-49ab-9866-0f9978fcdf1d`
4. Add new **Python Worker Extension** logs for reporting registered Python Worker Extensions.
5. Add unit test cases in **test_extension.py** to validate worker extension implementation.

## Description
This PR is based on https://github.com/Azure/azure-functions-python-worker/pull/757

This PR has a dependency on https://github.com/Azure/azure-functions-python-library/pull/81

This extension manager is intended to be used with azure.functions SDK. There're three roles involves in this scenarios.
1. Python function worker and SDK maintainers
2. Extension developers
3. Customers using Python Azure Functions

To enable this feature, two feature flags needs to be turned on in Python 3.6, 3.7 and 3.8.
Python 3.9 enables this feature by default.
**PYTHON_ISOLATE_WORKER_DEPENDENCIES** this will ensure when registering the extension, it will always direct to customer's azure-functions SDK instead of the built-in one.
**PYTHON_ENABLE_WORKER_EXTENSIONS** this will ensure the extension loaders are enabled. Otherwise, it will be ignored.

### 1. Python function worker and SDK maintainers
We decided to release a new extension module in azure.functions SDK where Extension developers can use it to register their extension. (https://github.com/Azure/azure-functions-python-library/pull/81)

### 2. Extension developers
Here are two examples for developing Application Level Extension and Function Level Extension for Python Worker.
Function Extension: https://github.com/Hazhzeng/functions-ext-profile
Application Extension: https://github.com/Hazhzeng/application-ext-timer

### 3. Azure Functions Users
The need to download the extension and azure-functions in their requirements.txt
```
azure-functions
git+https://github.com/Hazhzeng/functions-ext-profile
git+https://github.com/Hazhzeng/application-ext-timer

```
**To use a function extension:**
```python
import logging

import requests
from functions_ext_profile import ProfileExtension

profile_extension = ProfileExtension(__file__)


def main(req: func.HttpRequest) -> func.HttpResponse:
    logging.info('Start Executing HttpFuncExt')
    requests.get('https://microsoft.com')
    # profile_extension.property
    return 'HttpFuncExt Executed: called https://microsoft.com'

```
**To apply an application extension into every triggers:**
```python
import logging

import requests

# Import application level extension
from application_ext_timer import TimerExtension


def main(req: func.HttpRequest) -> func.HttpResponse:
    logging.info('Start Executing HttpFuncExt')
    requests.get('https://xbox.com')
    return 'HttpAppExt Executed: called https://xbox.com'

```

Fixes https://github.com/Azure/azure-functions-python-worker/issues/642
cc: @lzchen

---

### PR information


azure_functions_worker/logging.py
```diff
@@ -76,4 +76,4 @@ def enable_console_logging() -> None:
 def is_system_log_category(ctg: str) -> bool:
     # Category starts with 'azure_functions_worker' or
     # 'azure_functions_worker_errors' will be treated as system logs
-    return ctg.lower().startswith('azure_functions_worker')
+    return ctg.lower().startswith(SYSTEM_LOG_PREFIX)
```
vrdmr, line 79: Same question - `SYSTEM_LOG_PREFIX`, `CONSOLE_LOG_PREFIX` can be moved here. Better have the variables where they are used.

azure_functions_worker/dispatcher.py
```diff
@@ -255,8 +257,9 @@ async def _dispatch_grpc_request(self, request):
         self._grpc_resp_queue.put_nowait(resp)
 
     async def _handle__worker_init_request(self, req):
-        logger.info('Received WorkerInitRequest, request ID %s',
-                    self.request_id)
+        logger.info('Received WorkerInitRequest, '
+                    'python version %s, worker version %s, request ID %s',
+                    sys.version, __version__, self.request_id)
```
vrdmr, line 262: Good addition.

vrdmr (commented):
An initial set of small comments.

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,276 @@
...
+class ExtensionManager:
+    """This marks if the ExtensionManager has already proceeded a detection,
+    if so, the sdk will be cached in .extension_enabled_sdk
+    """
+    is_sdk_detected: bool = False
+
+    """This is a cache of azure.functions module that supports extension
+    interfaces. If this is None, that mean the sdk does not support extension.
+    """
+    extension_enabled_sdk: Optional[ModuleType] = None
```
vrdmr, line 37: Can `is_sdk_detected` and `extension_enabled_sdk` be private?

vrdmr: Also the methods such as `get_sdk_from_sys_path`, `is_extension_enabled_in_sdk`, `get_sdk_version` etc.? If yes, then please make it private as I don't see any other class needing access to these methods.

Hazhzeng: Yeah this is a nice catch, otherwise the extension manager is exposing too many interfaces.
I will only expose the essential ones `function_load_extension`, `invocation_extension`, `get_invocation_wrapper`, and `get_invocation_wrapper_async` and make all others private.

vrdmr (commented):
The second set of comments. Still to finish the `test_extension.py`

azure_functions_worker/dispatcher.py
Hazhzeng: Yes, we have a log statement in https://github.com/Azure/azure-functions-python-worker/pull/815/files#diff-073d7f68306c61100a03052187af456561bb52e007a2270f03aae14f4e0f1fc6R249, `Python Worker Extension is enabled in azure.functions (SDK version)` that we can query from the Kusto.

If adding the feature flag list here, this won't work for Linux Consumption as it requires placeholder environment reload. So I still prefer adding the log inside extension.py module is a better choice.

azure_functions_worker/dispatcher.py
```diff
@@ -373,20 +382,24 @@ async def _handle__invocation_request(self, req):
...
-                call_result = await fi.func(**args)
+                call_result = await self.__run_async_func(
+                    fi_context, fi.func, args
+                )
             else:
                 call_result = await self._loop.run_in_executor(
                     self._sync_call_tp,
-                    self.__run_sync_func, invocation_id, fi.func, args)
+                    self.__run_sync_func,
+                    invocation_id, fi_context, fi.func, args)
```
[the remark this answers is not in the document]
Hazhzeng: This dunder naming is from Elvis, the one who creates Python worker. I do agree changing it to single underscore will make the code cleaner.

The __run_sync_func will write to a thread local variable **_invocation_id_local**, this allows sync function to isolate their own invocation id inside the threadpool. I think what we can do here is to have **self._run_sync_func** to wrap the run_in_executor logic into one, and also have another **self._run_sync_func_internal** to access the thread local variable **_invocation_id_local**.

azure_functions_worker/dispatcher.py
```diff
@@ -304,6 +308,11 @@ async def _handle__function_load_request(self, req):
...
                 function_id, func, func_request.metadata)
 
+            ExtensionManager.function_load_extension(
+                function_name,
+                func_request.metadata.directory
+            )
+
             logger.info('Successfully processed FunctionLoadRequest, '
                         f'request ID: {self.request_id}, '
                         f'function ID: {function_id},'
```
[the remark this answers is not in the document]
Hazhzeng: > qq: If released - do we have a way to know how many function apps enable the PYTHON_ENABLE_WORKER_EXTENSIONS flag? Do we want to add another log message here to show the state of the features people are enabling?

Yes, we have `Python Worker Extension is enabled in azure.functions` logs to our Kusto FunctionsLogs table once this feature is enabled.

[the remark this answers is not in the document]
Hazhzeng: Yes we will log `Python Worker Extension Manager is loading {function_name}, {ExtensionList}` when the function is loaded.

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,278 @@
...
+class ExtensionManager:
+    """This marks if the ExtensionManager has already proceeded a detection,
+    if so, the sdk will be cached in ._extension_enabled_sdk
+    """
+    _is_sdk_detected: bool = False
+
+    """This is a cache of azure.functions module that supports extension
+    interfaces. If this is None, that mean the sdk does not support extension.
+    """
+    _extension_enabled_sdk: Optional[ModuleType] = None
```
vrdmr, line 39: the pydoc comment should go below the variable declarations. Not above. Same for the above `_is_sdk_detected` variable

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,278 @@
...
+            The folder path of the trigger
+            (e.g. /home/site/wwwroot/HttpTrigger).
+        """
+        sdk = cls._try_get_sdk_with_extension_enabled()
+        if sdk is None:
+            return
+        else:
+            cls._info_discover_extension_list(func_name, sdk)
+
+        # Invoke function hooks
```
vrdmr, line 69: This line isn't invoking the hooks, but gathering them. Can you clarify that?

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,278 @@
...
+        funcs = sdk.ExtensionMeta.get_function_hooks(func_name)
```
vrdmr, line 70: Also - if possible, can you list the expected datatype of funcs? is it a list or dict? If we cant define the exact type (as we don't have FuncExtensionHooks, etc.), then at least add comment.

vrdmr: Also, add a reference comment to where the implementation of get_function_hooks and get_application_hooks can be found.

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,278 @@
...
+        cls._safe_execute_function_load_hooks(
+            funcs, FUNC_EXT_POST_FUNCTION_LOAD, func_name, func_directory
+        )
+
+        # Invoke application hook
+        apps = sdk.ExtensionMeta.get_application_hooks()
+        cls._safe_execute_function_load_hooks(
+            apps, APP_EXT_POST_FUNCTION_LOAD, func_name, func_directory
+        )
```
vrdmr, line 79: I am assuming this is being called for each function load request - does that mean that this piece is going to be run for all the functions and result in the same?

vrdmr: Do you check if this has been processed before?

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,278 @@
...
+    @classmethod
+    @enable_feature_by(
+        flag=PYTHON_ENABLE_WORKER_EXTENSIONS,
+        flag_default=(
+            PYTHON_ENABLE_WORKER_EXTENSIONS_DEFAULT_39 if
+            is_python_version('3.9') else
+            PYTHON_ENABLE_WORKER_EXTENSIONS_DEFAULT
+        )
+    )
+    def invocation_extension(cls, ctx, hook_name, func_args, func_ret=None):
```
vrdmr, line 90: Ideally, this should be private as well - but I do see tests depending on/testing it directly - we can discuss if we need to focus more on the interfaces test or helpers as well (both have its importance)

azure_functions_worker/extension.py
Hazhzeng: Oh dude, thanks for pointing this out.

azure_functions_worker/extension.py
Hazhzeng: Great idea! To be honest, the reason why I was reluctant to add it is basically the azure.functions sdk is dynamically loaded. I forgot we can actually document the return types with comments as well. Adding it now, my bad.

azure_functions_worker/extension.py
Hazhzeng: Yes, so a single implementation in AppExtension.post_function_load_request_app_level will be applied to all functions when they get loaded.
image

azure_functions_worker/extension.py
Hazhzeng: Agree, I mark this as private field for now.

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        cls._invocation_extension(ctx, APP_EXT_PRE_INVOCATION, args)
+        cls._invocation_extension(ctx, FUNC_EXT_PRE_INVOCATION, args)
+        result = await function(**args)
+        cls._invocation_extension(ctx, FUNC_EXT_POST_INVOCATION, args, result)
+        cls._invocation_extension(ctx, APP_EXT_POST_INVOCATION, args, result)
+        return result
+
+    @staticmethod
+    def _get_sdk_from_sys_path() -> ModuleType:
```
vrdmr, line 151: Curious - could this be moved to utils? This doesn't feel like something that should be part of the ExtensionsManager - this is generic.

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,297 @@
...
+
+        Returns
+        -------
+        bool
+            True on azure.functions SDK supports extension registration
+        """
+        return getattr(module, 'ExtensionMeta', None) is not None
+
+    @staticmethod
+    def _get_sdk_version(module: ModuleType) -> str:
```
vrdmr, line 184: Same comment as above - this feels like something that should be outside of this specific class.

azure_functions_worker/extension.py
```diff
@@ -0,0 +1,297 @@
...
+        logger.warning(
+            f'The azure.functions ({cls._get_sdk_version(sdk)}) does not '
+            'support Python worker extensions. If you believe extensions '
+            'are correctly installed, please set the '
+            f'{PYTHON_ISOLATE_WORKER_DEPENDENCIES} and '
+            f'{PYTHON_ENABLE_WORKER_EXTENSIONS} to "true"'
+        )
+
+    @classmethod
+    def _warn_context_has_no_function_name(cls):
```
vrdmr, line 293: I didn't see any usage of this method. Can you double check?

vrdmr (approved):
LGTM;

minor comments.

azure_functions_worker/extension.py
Hazhzeng: Good catch, I will remove this.

PR #822, opened 2021-02-19
Prevent worker isolation from breaking apps without sdk in .python_packages

Hazhzeng:
## Description
Fix CI Build Error: https://azure-functions.visualstudio.com/azure-functions-docker/_build/results?buildId=7870&view=results

The root cause of this is because the [Python Function App testcase](https://github.com/Azure/azure-functions-docker/blob/dev/test/index.ts#L22) does not have **azure-functions** module in **.python_packages**. [python-functions.zip](https://github.com/Azure/azure-functions-python-worker/files/6008596/python-functions.zip)

This is not a valid function app for Azure Functions v3 but it was when we GA Python on Linux Consumption. Previously, we **hide azure-functions away from requirements.txt** and only place it along with azure-functions-worker packages. So even without the .python_packages, the function app should still load.

This is fixed by adding the worker dependencies as the last entry in sys.path when switching to customer dependencies.

---

### PR information


azure_functions_worker/dispatcher.py
```diff
@@ -264,7 +264,7 @@ async def _handle__worker_init_request(self, req):
         }
 
         # Can detech worker packages
-        DependencyManager.use_customer_dependencies()
+        DependencyManager.prioritize_customer_dependencies()
```
vrdmr, line 267: Good naming! :)

azure_functions_worker/utils/dependency.py
```diff
@@ -200,6 +206,12 @@ def reload_all_namespaces_from_customer_deps(cls, cx_working_dir: str):
         cls._remove_from_sys_path(cls.worker_deps_path)
         cls._add_to_sys_path(cls.cx_deps_path, True)
         cls._add_to_sys_path(working_directory, False)
+
+        # Deprioritize worker dependencies but don't completely remove it
+        # Otherwise, it will break some really old function apps, those
+        # don't have azure-functions module in .python_packages
+        cls._add_to_sys_path(cls.worker_deps_path, False)
```
vrdmr, line 213: If possible - add the PR number in both comments (one above as well). That'll redirect people to the PR/issue which would have more verbose details of the issue.

tests/unittests/test_utilities_dependency.py
```diff
@@ -258,6 +258,41 @@ def test_add_to_sys_path_importer_cache_reloaded(self):
             os.path.join(self._customer_deps_path, 'common_module')
         )
 
+    def test_reload_all_namespaces_from_customer_deps(self):
```
vrdmr, line 261: Can you please add a small description of what this test is doing?

vrdmr (approved):

PR #844, opened 2021-04-28
Using the cache for function bindings and not deleting shared memory resources upon request to close

gohar94:
## Description
When the host sends a message to close shared memory resources, instead of deleting the shared memory resources (i.e. files in the case of Linux) just drop the handle and let the worker delete the resources.

When the host advertises the capability for using FunctionDataCache, use that even for small objects (without imposing the minimum size constraint.)

This PR is part of the work described in Azure/azure-functions-host#7310

---

### PR information


azure_functions_worker/bindings/meta.py
```diff
@@ -111,6 +111,14 @@ def get_datum(binding: str, obj: typing.Any,
     return datum
 
 
+def is_cache_supported(datum: datumdef.Datum):
+    if datum.type == 'bytes':
+        return True
+    elif datum.type == 'string':
```
goiri, line 117: you can do if without else

azure_functions_worker/bindings/meta.py
```diff
@@ -120,13 +128,16 @@ def to_outgoing_proto(binding: str, obj: typing.Any, *,
...
-                              shmem_mgr) \
+                              shmem_mgr,
+                              is_function_data_cache_enabled: bool) \
         -> protos.ParameterBinding:
     datum = get_datum(binding, obj, pytype)
     shared_mem_value = None
     # If shared memory is enabled and supported for the given datum, try to
     # transfer to host over shared memory as a default
-    if shmem_mgr.is_enabled() and shmem_mgr.is_supported(datum):
+    can_transfer_over_shmem = shmem_mgr.is_supported(datum) or \
```
goiri, line 138: Let's put this whole thing condition in a function?

azure_functions_worker/bindings/shared_memory_data_transfer/shared_memory_manager.py
```diff
@@ -170,7 +171,7 @@ def free_mem_map(self, mem_map_name: str) -> bool:
                 f'Cannot find memory map in list of allocations {mem_map_name}')
             return False
         shared_mem_map = self.allocated_mem_maps[mem_map_name]
-        success = shared_mem_map.dispose()
+        success = shared_mem_map.dispose(is_delete_backing_resources)
```
vrdmr, line 174: This flag if it is disposed of or not. Correct?

vrdmr: Can you have a better naming for this? The "is_delete" part was confisuing. It is more like "to_delete..."?

azure_functions_worker/dispatcher.py
```diff
@@ -262,6 +263,12 @@ async def _handle__worker_init_request(self, req):
                     'python version %s, worker version %s, request ID %s',
                     sys.version, __version__, self.request_id)
 
+        worker_init_request = req.worker_init_request
+        host_capabilities = worker_init_request.capabilities
```
vrdmr, line 267: Nice. I think this is useful to be logged as well in the future. I'll take a note of this

vrdmr (commented):
minor comments. Also, can you please add testing for cache-enabled scenarios? to see if the same shared mem is reused or not?

azure_functions_worker/constants.py
```diff
@@ -10,6 +10,7 @@
 RPC_HTTP_TRIGGER_METADATA_REMOVED = "RpcHttpTriggerMetadataRemoved"
 WORKER_STATUS = "WorkerStatus"
 SHARED_MEMORY_DATA_TRANSFER = "SharedMemoryDataTransfer"
+FUNCTION_DATA_CACHE = "FunctionDataCache"
```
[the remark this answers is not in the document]
gohar94: Yes - [here](https://github.com/Azure/azure-functions-host/blob/4a50ecf75833e2ab7a1392ff6e726ce34d28b56a/src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs#L47) is the corresponding flag in the host.

gohar94:
> minor comments. Also, can you please add testing for cache-enabled scenarios? to see if the same shared mem is reused or not?

That's a very good point.
The problem right now is that the caching piece will not be enabled until one of the azure storage extension [PR](https://github.com/Azure/azure-sdk-for-net/pull/20209)s is checked in. That PR will not be checked in until the azure-webjobs-SDK [PR](https://github.com/Azure/azure-webjobs-sdk/pull/2692) package is published to nuget.org - this will happen in a few weeks as part of some deployment cycle which I don't fully understand. Once that is checked in, then we can have a subsequent PR here for the Python worker which consumes the latest host and storage extension in the tests so that we can put the tests you mentioned. Does that sounds like a fair plan?

azure_functions_worker/dispatcher.py
```diff
@@ -528,7 +537,7 @@ async def _handle__close_shared_memory_resources_request(self, req):
         try:
             for mem_map_name in map_names:
                 try:
-                    success = self._shmem_mgr.free_mem_map(mem_map_name)
+                    success = self._shmem_mgr.free_mem_map(mem_map_name, False)
```
goiri, line 540: For readability, it might be better to put this False in a variable with a name or to do the full param: 
to_delete_backing_resources: False

azure_functions_worker/bindings/meta.py
```diff
@@ -111,6 +111,14 @@ def get_datum(binding: str, obj: typing.Any,
     return datum
 
 
+def is_cache_supported(datum: datumdef.Datum):
```
vrdmr, line 114: Could you please clarify here that we are checking if the caching a datatype is supported or not. More like `does_datatype_support_caching`? As I think this confuses with `is_function_data_cache_enabled` and only looking at the definition of this function specifies why this is required.

Also, could you please make it private?

vrdmr: Also - why not look at `in`

```python
>>> x = ("string", "bytes")
>>> a = "bytes"
>>> b = "blah"
>>> a in x
True
>>> b in x
False
```

azure_functions_worker/bindings/meta.py
```diff
@@ -111,6 +111,14 @@ def get_datum(binding: str, obj: typing.Any,
...
+    if datum.type == 'bytes':
+        return True
+    elif datum.type == 'string':
+        return True
+    return False
```
vrdmr, line 119:
```suggestion
    supported_datatypes = ('bytes', 'string')
    return datum.type in supported_datatypes
```

azure_functions_worker/dispatcher.py
```diff
@@ -526,12 +535,13 @@ async def _handle__close_shared_memory_resources_request(self, req):
         results = {mem_map_name: False for mem_map_name in map_names}
 
         try:
-            for mem_map_name in map_names:
+            for map_name in map_names:
                 try:
-                    success = self._shmem_mgr.free_mem_map(mem_map_name)
-                    results[mem_map_name] = success
+                    to_delete = False
```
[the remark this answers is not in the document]
vrdmr: Or this should also check that app setting.

vrdmr (changes requested):

azure_functions_worker/dispatcher.py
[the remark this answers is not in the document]
gohar94: This is a great catch - my intention was to change the host to handle the cache/non-cache case but I missed that. For now, I will just check the app setting. Tested manually on WSL Linux and seems to be fine (and indeed it was not fine before this) - thanks for picking up on this!

azure_functions_worker/bindings/meta.py
gohar94: Accepted your suggested change.

azure_functions_worker/bindings/meta.py
goiri: The condition for this could still be made more readable.
I think adding all in a single function is cleaner.
```
def function_name(XXXX):
  if shmem_mgr.is_enabled():
    return False
  if not is_function_data_cache_enabled:
    return False
  if not _does_datatype_support_caching(datum):
    return False
  if not shmem_mgr.is_supported(datum):
    return False
  return True
```

azure_functions_worker/dispatcher.py
```diff
@@ -526,12 +535,22 @@ async def _handle__close_shared_memory_resources_request(self, req):
...
+                    if self._function_data_cache_enabled:
+                        # If the cache is enabled, let the host decide when to
+                        # delete the resources.
+                        # Just drop the reference from the worker.
+                        to_delete = False
+                    else:
+                        # If the cache is not enabled, the worker should free
+                        # the resources as at this point the host has read the
+                        # memory maps and does not need them.
+                        to_delete = True
```
vrdmr, line 549:
```suggestion
                    to_delete_reference = False if self._function_data_cache_enabled else True
```

And please add the comment in the python docs section of this method itself.

azure_functions_worker/dispatcher.py
```diff
@@ -526,12 +535,22 @@ async def _handle__close_shared_memory_resources_request(self, req):
...
+                    success = self._shmem_mgr.free_mem_map(map_name, to_delete)
```
vrdmr, line 550:
```suggestion
                    success = self._shmem_mgr.free_mem_map(map_name, to_delete_reference)
```

If above suggestion is included, please include this and also check if there is any other place `to_delete` is mentioned.

vrdmr (approved):
nit changes. otherwise, LGTM. 🚢

vrdmr (approved):

Issue #840, opened 2021-04-05
Add logging when wsgi library is initiated and executed


stefanushinardi:
cc: @Hazhzeng

Hazhzeng:
Deployment in progress. This feature will be shipped with the host version 3.1.2

vrdmr:
Shipped as part of V3 and V4 deployments. Closing this issue.

PR #850, opened 2021-05-12
Categorize SDK logs as system logs

Hazhzeng:
## Description

Dependency https://github.com/Azure/azure-functions-python-library/pull/88
Fixes #https://github.com/Azure/azure-functions-python-worker/issues/840

From now on, we will treat 'azure.functions' namespace and its child namespace as system logs.
After merging this PR, we should be able to see the following log lines in our Kusto table:
1. Instantiating Azure Functions ASGI middleware.
2. Instantiating Azure Functions WSGI middleware.

This will help the PM team calculate the usage on WSGI/ASGI middlewares.

Core changes in `azure_functions_worker/logging.py`.

cc: @stefanushinardi @vrdmr @AnatoliB 

---

### PR information

### DIFF
--- a/azure_functions_worker/logging.py
+++ b/azure_functions_worker/logging.py
@@ -9,6 +9,7 @@ import sys
 # Logging Prefixes
 CONSOLE_LOG_PREFIX = "LanguageWorkerConsoleLog"
 SYSTEM_LOG_PREFIX = "azure_functions_worker"
+SDK_LOG_PREFIX = "azure.functions"
 
 
 logger: logging.Logger = logging.getLogger('azure_functions_worker')
@@ -76,6 +77,11 @@ def enable_console_logging() -> None:
 
 
 def is_system_log_category(ctg: str) -> bool:
-    # Category starts with 'azure_functions_worker' or
-    # 'azure_functions_worker_errors' will be treated as system logs
-    return ctg.lower().startswith(SYSTEM_LOG_PREFIX)
+    # Category starts with the following name will be treated as system logs,
+    # and send to our customers and function teams Kusto logs:
+    # 1. 'azure_functions_worker' (Worker Info)
+    # 2. 'azure_functions_worker_errors' (Worker Error)
+    # 3. 'azure.functions' (SDK)
+    ctg_lower = ctg.lower()
+    return ctg_lower.startswith(SYSTEM_LOG_PREFIX) or (
+        ctg_lower.startswith(SDK_LOG_PREFIX))
--- a/azure_functions_worker/logging.py
+++ b/azure_functions_worker/logging.py
@@ -77,11 +77,15 @@ def enable_console_logging() -> None:
 
 
 def is_system_log_category(ctg: str) -> bool:
-    # Category starts with the following name will be treated as system logs,
-    # and send to our customers and function teams Kusto logs:
-    # 1. 'azure_functions_worker' (Worker Info)
-    # 2. 'azure_functions_worker_errors' (Worker Error)
-    # 3. 'azure.functions' (SDK)
-    ctg_lower = ctg.lower()
-    return ctg_lower.startswith(SYSTEM_LOG_PREFIX) or (
-        ctg_lower.startswith(SDK_LOG_PREFIX))
+    """Check if the logging namespace belongs to system logs. Category starts
+    with the following name will be treated as system logs.
+    1. 'azure_functions_worker' (Worker Info)
+    2. 'azure_functions_worker_errors' (Worker Error)
+    3. 'azure.functions' (SDK)
+
+    Expected behaviors for sytem logs and customer logs are listed below:
+                  local_console  customer_app_insight  functions_kusto_table
+    system_log    false          false                 true
+    customer_log  true           true                  false
+    """
+    return ctg.startswith(SYSTEM_LOG_PREFIX) or ctg.startswith(SDK_LOG_PREFIX)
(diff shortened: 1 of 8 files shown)


tests/unittests/log_filtering_functions/sdk_logging/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+import logging
+
+import azure.functions as func
+
+
+sdk_logger = logging.getLogger('azure.functions')
```
vrdmr, line 8: So - if someone creates a new logger using `logging.getLogger('azure.functions')` - all the logs logged would now be send to us.

Does this also send to AppInsights? I was assuming no but then I read this - 

```
Here's a list of expected behaviors:
                  local_console  customer_app_insight  functions_kusto_table
    system_log    false          true                  true
    customer_log  true           true                  false
```

Do we send all the system logs to AppInsights. I don't think so - but I'll confirm as well.

vrdmr: And just doing `logging.info` would still log to App Insights. Right?

tests/unittests/test_log_filtering_functions.py
```diff
@@ -61,4 +70,40 @@ def check_log_debug_with_user_logging(self, host_out: typing.List[str]):
...
+        self.assertEqual(r.text, 'OK-sdk-submodule-logging')
+
+    def check_log_info_with_sdk_submodule_logging(self,
+                                                  host_out: typing.List[str]):
+        # See TestLogFilteringFunctions docstring
+        # System log should not be captured in console
+        self.assertNotIn('sdk_submodule_logger info', host_out)
+        self.assertNotIn('sdk_submodule_logger warning', host_out)
+        self.assertNotIn('sdk_submodule_logger error', host_out)
+        self.assertNotIn('sdk_submodule_logger debug', host_out)
```
vrdmr, line 109: Stretch: Any way to check that AppInsights logs as well - if it is being logged?

azure_functions_worker/logging.py
```diff
@@ -76,6 +77,11 @@ def enable_console_logging() -> None:
 
 
 def is_system_log_category(ctg: str) -> bool:
-    # Category starts with 'azure_functions_worker' or
-    # 'azure_functions_worker_errors' will be treated as system logs
-    return ctg.lower().startswith(SYSTEM_LOG_PREFIX)
+    # Category starts with the following name will be treated as system logs,
```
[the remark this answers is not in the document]
Hazhzeng: @vrdmr, this is a very good point, and I don't even think we need to use the ctg.lower(). In the official [Python logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#using-logging-in-multiple-modules), the record.name is actually case sensitive. In the code snippet they share below, the handler is not calling lower() to the record.name.

```python
def handleLogRecord(self, record):
    # if a name is specified, we use the named logger rather than the one
    # implied by the record.
    if self.server.logname is not None:
        name = self.server.logname
    else:
        name = record.name
    logger = logging.getLogger(name)
```

tests/unittests/log_filtering_functions/sdk_logging/__init__.py
Hazhzeng: Yeah, I will make an E2E test and see if the application insights logs can be checked efficiently. As I know, the application logs require 30 seconds delay to get the log statement.

Hazhzeng: Just checked the application insights manually, this is a good catch. Seems like the system_log does not go to the customer_app_insight, I will amend the PR now.

image

tests/unittests/test_log_filtering_functions.py
Hazhzeng: Yeah, I actually tried to push traces to application insights, and fetch back the result from the GET https://api.applicationinsights.io/v1/apps/ API. The trace exists in Live Metrics tab, but they do not show up in the log analytics and cannot be queried from the REST API endpoint. So I think we will skip that.

tests/unittests/test_log_filtering_functions.py
```diff
@@ -20,6 +19,15 @@
 
 
 class TestLogFilteringFunctions(testutils.WebHostTestCase):
+    """This class is for testing the logger behavior in Python Worker when
+    dealing with customer's log and system's log. Here's a list of expected
+    behaviors:
```
vrdmr, line 24: nit; you don't need to mention this - if you have mentioned this earlier. Just give a reference of `is_system_log_category` and that should be good enough.

tests/unittests/test_log_filtering_functions.py
```diff
@@ -50,6 +58,7 @@ def check_log_debug_logging(self, host_out: typing.List[str]):
         self.assertIn('logging info', host_out)
         self.assertIn('logging warning', host_out)
         self.assertIn('logging error', host_out)
+        # See HOST_JSON_TEMPLATE_WITH_LOGLEVEL_INFO, debug log is disabled
```
vrdmr, line 61: Sorry - didnt understand this - where is this pointing to?

tests/unittests/test_logging.py
```diff
@@ -0,0 +1,35 @@
...
+
+from azure_functions_worker import logging as flog
+
+
+class TestLogging(unittest.TestCase):
+    """This class is for testing the grpc logging behavior in Python Worker.
+    Here's a list of expected behaviors:
+                  local_console  customer_app_insight  functions_kusto_table
+    system_log    false          false                 true
+    customer_log  true           true                  false
```
vrdmr, line 13: same as above

vrdmr (approved):
nit comments; lgtm.

PR #875, opened 2021-07-20
Update the max TP threads count to sys.maxsize

### LINKED ISSUE #873: Allow specifying PYTHON_THREADPOOL_THREAD_COUNT larger than 32
Currently, the worker limits the maximum value of `PYTHON_THREADPOOL_THREAD_COUNT` to 32. Any value larger than that is ignored. There is no good reason for this in general case: if someone/something decided a larger number works better for a specific use case, this should be respected.

vrdmr:
## Description
With new work being done on the Host side control concurrency, the 32 thread number isn't required. Setting the max value to 1024.

Fixes #873

---

### PR information


azure_functions_worker/dispatcher.py
```diff
@@ -570,7 +570,9 @@ def tp_max_workers_validator(value: str) -> bool:
             if int_value < PYTHON_THREADPOOL_THREAD_COUNT_MIN or (
                     int_value > PYTHON_THREADPOOL_THREAD_COUNT_MAX):
```
AnatoliB, line 571: I still don't understand why we even have this check at all. Regardless of the MAX value, what is the reason?

As a user, I expect the value I specify to be either accepted (preferred) or at least explicitly rejected (if there is an actual reason). Right now, if I specify 1025, this value will not be accepted and, even worse, the worker will fall back to the very low default, and I may not notice the warning. And, _in order to achieve these troubles_, we have quite a lot of code!

If we didn't have this check at all, what would be the motivation for us to add it? Are we assuming we know more than the user who specified a certain value? How can we know that?

Hazhzeng (approved):

azure_functions_worker/dispatcher.py
vrdmr: After offline conversation - we have decided to remove any artificial limits on the maximum number of threads.

AnatoliB (approved):
I would even recommend _removing_ the code performing this check, including the unit tests and comments, as this is effectively dead code now. But I'll leave this up to you. Other than that, LGTM and thank you!

vrdmr:
@AnatoliB - Yes, that's the plan - to remove the code and the unit tests. I wanted to get this change in sooner to unblock the perf testing work and get this part of the next week's release.

PR #890, opened 2021-09-14
Setting special limits for Python 3.[6|7]

### LINKED ISSUE #873: Allow specifying PYTHON_THREADPOOL_THREAD_COUNT larger than 32

vrdmr:
## Description
For Python 3.6 and 3.7, setting max workers to large values is problematic as the implementation of ThreadPoolExecutor for both that versions do not reuse threads until max_workers number is hit, and that is a danger to the functioning of the app. This PR limits the number of max_workers to 32 for 3.6 and 3.7 and allows it to be `sys.maxint` for Python 3.8+.

Reference doc - [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)

> Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.
> 
> ThreadPoolExecutor now reuses idle worker threads before starting max_workers worker threads too.

Fixes #873

---

### PR information


azure_functions_worker/dispatcher.py
```diff
@@ -567,25 +567,28 @@ def tp_max_workers_validator(value: str) -> bool:
...
-                               'Reverting to default value for max_workers')
+                               'sys.maxint. Reverting to default value for '
+                               'max_workers')
                 return False
-
             return True
 
         # Starting Python 3.9, worker won't be putting a limit on the
         # max_workers count in the created threadpool.
         default_value = None if sys.version_info.minor == 9 \
```
AnatoliB, line 581: Shouldn't this be `sys.version_info.minor >= 9`?

AnatoliB (approved):
LGTM

azure_functions_worker/dispatcher.py
vrdmr: Currently, there isn't any 3.10, but yes, this would be updated whenever the work for 3.10 is done. As we don't know how the implementation of ThreadPoolExecutor is and what setting default value would have an impact on. Will keep that in mind.

PR #939, opened 2022-01-12
Enable debug logging using the PYTHON_ENABLE_DEBUG_LOGGING flag

YunchuWang:
## Description
Per customers' requests to log messages at debug level in their python function code and view them in App Insights, there will be a new field named PYTHON_ENABLE_DEBUG_LOGGING in Function App Settings supported. It will determine whether debug level logs in the customer function code will be recorded to App Insights. When the field is set to 0 or unspecified in the App Settings, debug level logs will not be recorded; when the field is set to 1, debug level logs will be recorded.

Fixes #

---

### PR information


azure_functions_worker/dispatcher.py
```diff
@@ -146,7 +147,8 @@ async def dispatch_forever(self):
             root_logger = logging.getLogger()
 
             # Don't change this unless you read #780 and #745
```
vrdmr, line 149: You can remove this comment - or update it with some more context.

azure_functions_worker/dispatcher.py
```diff
@@ -146,7 +147,8 @@ async def dispatch_forever(self):
...
-            root_logger.setLevel(logging.INFO)
+            root_logger.setLevel(logging.DEBUG if is_envvar_true(
+                PYTHON_ENABLE_DEBUG_LOGGING) else logging.INFO)
```
vrdmr, line 151: I would recommend keeping the more common case as default. Also - it would be a little more readable to do something like the following:

```suggestion
            log_level = logging.INFO if not is_envvar_true(
                PYTHON_ENABLE_DEBUG_LOGGING) else logging.DEBUG
            root_logger.setLevel(log_level)
```

azure_functions_worker/dispatcher.py
```diff
@@ -484,6 +486,11 @@ async def _handle__function_environment_reload_request(self, req):
                 self._create_sync_call_tp(self._get_sync_tp_max_workers())
             )
 
+            # Apply PYTHON_ENABLE_DEBUG_LOGGING
+            if is_envvar_true(PYTHON_ENABLE_DEBUG_LOGGING):
+                root_logger = logging.getLogger()
+                root_logger.setLevel(logging.DEBUG)
+
```
vrdmr, line 493: The comment doesn't add much help. You can remove this.

azure_functions_worker/main.py
```diff
@@ -12,12 +12,11 @@ def parse_args():
...
                         help="host address")
     parser.add_argument('--port', type=int,
-                        help='id for the requests')
+                        help='port number')
     parser.add_argument('--workerId', dest='worker_id',
                         help='id for the worker')
     parser.add_argument('--requestId', dest='request_id',
-                        help='log destination: stdout, stderr, '
-                             'syslog, or a file path')
+                        help='id of the request')
```
vrdmr, line 19: Good catch.

tests/endtoend/test_linux_consumption.py
```diff
@@ -1,4 +1,4 @@
-from unittest import TestCase, skip
+from unittest import TestCase
```
vrdmr, line 1: Can you copy and put the copyright comment (from any other file) to this file as well?

tests/unittests/test_enable_debug_logging_functions.py
```diff
@@ -0,0 +1,68 @@
...
+# Licensed under the MIT License.
+import typing
+import os
+from unittest.mock import patch
+
+from azure_functions_worker import testutils
+from azure_functions_worker.constants import PYTHON_ENABLE_DEBUG_LOGGING
+
+
+class TestDebugLoggingEnabledFunctions(testutils.WebHostTestCase):
```
vrdmr, line 11: Checking - didn't you have to update the host.json file for enabling the logging? Or is it enabled by default now?

If yes - Do we need another test where it is not enabled by default and we are still enabling the flag and checking that the new debug log doesn't get logged?

Ideally - there is a potential issue - 
If a customer didn't update host.json, but enables the new feature flag, we start sending the logs from our end. This could result the choking of the gRPC channel that we already know about. If we can also get the host.json information from the host (which I think we do get during worker_init), then we can also do some check if we need to actually send the logs at all or not.
*Note*: Not very important but something to be aware of.

vrdmr (changes requested):
minor changes requested. Overall looks good. :)

tests/unittests/test_enable_debug_logging_functions.py
YunchuWang: Good catch, will add a test case for that. As discussed, the host.json logging filter is set at function level while this flag is at function app level for the shared root logger. It is not possible to define the flag validity if functions have inconsistent log levels in the host.json. Separate investigation/change will be made to warn customers in case of inconsistent host.json log level and PYTHON_ENABLE_DEBUG_LOGGING value being set.

azure_functions_worker/dispatcher.py
YunchuWang: will do

azure_functions_worker/dispatcher.py
```diff
@@ -145,8 +146,9 @@ async def dispatch_forever(self):
             logging_handler = AsyncLoggingHandler()
             root_logger = logging.getLogger()
 
-            # Don't change this unless you read #780 and #745
-            root_logger.setLevel(logging.INFO)
+            log_level = logging.INFO if not is_envvar_true(
```
gavin-aguiar, line 149: You can use the is_envvar_false method instead of the not is_envvar_true

tests/unittests/enable_debug_logging_functions/debug_logging/main.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
gavin-aguiar, line 1: Why add this function? Can we just reuse the functions in log_filtering_functions directory?

tests/unittests/test_http_functions.py
```diff
@@ -98,34 +97,22 @@ def check_log_async_logging(self, host_out: typing.List[str]):
         self.assertIn('hello info', host_out)
         self.assertIn('and another error', host_out)
 
-    @unittest.skip("Reverting the debug logs PR as host currently cannot handle"
-                   "apps with lot of debug statements. Reverting PR: "
-                   "azure-functions-python-worker/pull/745")
     def test_debug_logging(self):
```
gavin-aguiar, line 100: Do we need all of these tests if you are already testing this in the test_enable_debug_logging_functions.py? If not, then I guess these can be removed

YunchuWang: yes, this is kept for testing flag unset case

azure_functions_worker/dispatcher.py
YunchuWang: i originally thought too but it fails my tests because this method returns False if env key is unset.

```
def is_envvar_false(env_key: str) -> bool:
    if os.getenv(env_key) is None:
        return False

    return is_false_like(os.environ[env_key])
```

tests/unittests/enable_debug_logging_functions/debug_logging/main.py
YunchuWang: good catch, will remove

tests/endtoend/test_linux_consumption.py
```diff
@@ -134,6 +135,59 @@ def test_old_protobuf(self):
             self.assertIn('grpc', content)
             self.assertEqual(resp.status_code, 200)
 
+    def test_debug_logging_disabled(self):
+        """An HttpTrigger function app with 'azure-functions' library
+        should return 200 and by default customer debug logging should be
+        disabled.
+        """
+        with LinuxConsumptionWebHostController("3", self._py_version) as ctrl:
```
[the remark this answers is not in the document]
YunchuWang: Adding to the question, checked all test_linux_consumption test cases are testing v3 in specific, setting to other host versions in tests returns cant find images error. If we are testing v4, we need to refactor test_lc and testutils_lc files?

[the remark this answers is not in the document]
YunchuWang: As discussed, will create a separate task to track for refactoring/creating host V4 tests. Still using V3 for this test.

tests/unittests/test_enable_debug_logging_functions.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertEqual(r.text, 'OK-debug')
+
+    def check_log_debug_logging_enabled(self, host_out: typing.List[str]):
+        self.assertIn('logging info', host_out)
+        self.assertIn('logging warning', host_out)
+        self.assertIn('logging debug', host_out)
+        self.assertIn('logging error', host_out)
+
+
+class TestDebugLoggingDisabledFunctions(testutils.WebHostTestCase):
```
vrdmr, line 54: Can you add PyDoc comments to describe what the tests do? That'll help in the future.

tests/unittests/test_enable_debug_logging_functions.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertEqual(r.text, 'OK-debug')
+
+    def check_log_debug_logging_disabled(self, host_out: typing.List[str]):
+        self.assertIn('logging info', host_out)
+        self.assertIn('logging warning', host_out)
+        self.assertIn('logging error', host_out)
+        self.assertNotIn('logging debug', host_out)
+
+
+class TestLogFilteringFunctions(testutils.WebHostTestCase):
```
vrdmr, line 84: Can you please change the class name to describe what this test is specifically testing?

vrdmr: Also the same comments of comments.

vrdmr (approved):
LGTM.

Left minor comments.

PR #978, opened 2022-03-08
Added an option in setup.py to use any branch from azure-functions-host

gavin-aguiar:
Added an option in setup.py to use any branch from azure-functions-host
Usage

python setup.py webhost
python setup.py webhost --use-branch=dev
python setup.py webhost --use-branch=abc/branch

## Description

Fixes #

---

### PR information

- [x ] I have read the [contribution guidelines](https://github.com/Azure/azure-functions-python-worker/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ from azure_functions_worker.version import VERSION
 # The GitHub repository of the Azure Functions Host
 WEBHOST_GITHUB_API = "https://api.github.com/repos/Azure/azure-functions-host"
 WEBHOST_TAG_PREFIX = "v4."
+WEBHOST_GIT_REPO = "https://github.com/Azure/azure-functions-host/archive"
 
 # Extensions necessary for non-core bindings.
 AZURE_EXTENSIONS = """\
@@ -56,7 +57,6 @@ AZURE_EXTENSIONS = """\
 </Project>
 """
 
-
 NUGET_CONFIG = """\
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
@@ -75,7 +75,6 @@ NUGET_CONFIG = """\
 </configuration>
 """
 
-
 CLASSIFIERS = [
     "Development Status :: 5 - Production/Stable",
     'Programming Language :: Python',
@@ -91,7 +90,6 @@ CLASSIFIERS = [
     "Intended Audience :: Developers",
 ]
 
-
 PACKAGES = [
     "azure_functions_worker",
     "azure_functions_worker.protos",
@@ -103,14 +101,12 @@ PACKAGES = [
     "azure_functions_worker._thirdparty"
 ]
 
-
 INSTALL_REQUIRES = [
     "grpcio~=1.43.0",
     "grpcio-tools~=1.43.0",
     "protobuf~=3.19.3",
     "azure-functions==1.9.0"
 ]
-
 
 EXTRA_REQUIRES = {
     "dev": [
@@ -135,6 +131,7 @@ EXTRA_REQUIRES = {
 
 class BuildGRPC:
     """Generate gRPC bindings."""
+
     def _gen_grpc(self):
         root = pathlib.Path(os.path.abspath(os.path.dirname(__file__)))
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -275,16 +275,19 @@ def run(self):
 class Webhost(distutils.cmd.Command):
     description = 'Download and setup Azure Functions Web Host.'
     user_options = [
-        ('webhost-version', None,
+        ('webhost-version=', None,
             'A Functions Host version to be downloaded (e.g. 3.0.15278).'),
-        ('webhost-dir', None,
+        ('webhost-dir=', None,
             'A path to the directory where Azure Web Host will be installed.'),
+        ('use-branch=', None,
```
YunchuWang, line 282: use-branch sounds like a boolean field, maybe branch-name?

setup.py
```diff
@@ -410,9 +421,11 @@ def _compile_webhost(webhost_dir: pathlib.Path):
 
     def run(self):
         # Prepare webhost
-        zip_path = self._download_webhost_zip(self.webhost_version)
+        zip_path = self._download_webhost_zip(self.webhost_version,
+                                              self.use_branch)
         self._create_webhost_folder(self.webhost_dir)
-        self._extract_webhost_zip(version=self.webhost_version,
+        version = self.use_branch or self.webhost_version
+        self._extract_webhost_zip(version=version.replace('/', '-'),
```
YunchuWang, line 428: add null check in case version is None?

setup.py
```diff
@@ -439,7 +452,10 @@ def run(self) -> None:
         for dir_to_delete in self.dir_list_to_delete:
             dir_delete = pathlib.Path(dir_to_delete)
             if dir_delete.exists():
-                dir_util.remove_tree(str(dir_delete))
+                try:
+                    shutil.rmtree(dir_delete)
+                except:
+                    print('Error deleting directory')
```
YunchuWang, line 458: Log directory to delete

vrdmr, line 458: if possible, print the file that it failed to delete.

setup.py
```diff
@@ -313,14 +316,22 @@ def _get_webhost_version() -> str:
...
         temporary_file = tempfile.NamedTemporaryFile()
-        zip_url = (
-            'https://github.com/Azure/azure-functions-host/archive/'
-            f'v{version}.zip'
-        )
-        print(f'Downloading Functions Host ({version}) from {zip_url}')
+
+        if branch is not None:
+            zip_url = (
+                'https://github.com/Azure/azure-functions-host/archive/'
```
vrdmr, line 325: Minor thing: Can you Constantify the first part of the URL?

setup.py
```diff
@@ -275,16 +275,19 @@ def run(self):
...
     description = 'Download and setup Azure Functions Web Host.'
     user_options = [
-        ('webhost-version', None,
+        ('webhost-version=', None,
             'A Functions Host version to be downloaded (e.g. 3.0.15278).'),
-        ('webhost-dir', None,
+        ('webhost-dir=', None,
             'A path to the directory where Azure Web Host will be installed.'),
+        ('use-branch=', None,
+            'A branch from where azure-functions-host will be installed')
```
vrdmr, line 283: Does `distutils.cmd.Command` support example usages section?

vrdmr (approved):

setup.py
gavin-aguiar: We dont need a null check here.. If branch_name is none then it should use webhost_version which can never be none since we default it to some value

setup.py
arroyc: nit: make the comments similar, for example webhost-version you have an e.g. 3.0.15278, it will be nice if you provide e.g for webhost-dir and branch-name/use-branch as well

setup.py
arroyc: +1

setup.py
```diff
@@ -313,14 +316,22 @@ def _get_webhost_version() -> str:
...
+
+        if branch is not None:
+            zip_url = (
+                'https://github.com/Azure/azure-functions-host/archive/'
+                f'refs/heads/{branch}.zip'
+            )
+        else:
+            zip_url = (
+                'https://github.com/Azure/azure-functions-host/archive/'
+                f'v{version}.zip'
```
arroyc, line 331: nit: ditto for url constant

setup.py
arroyc: do we have a logger where we are flashing all these logs or the standard out gets piped to somewhere else by default

PR #1039, opened 2022-05-23
Add dev container

tonybaloney:
Makes it easier for people to contribute to the project and get a dev environment working.

Does the following:

- Creates a Python 3.9 environment
- Installs the web host and .NET 6 SDK
- Installs the project development requirements
- Configures VS Code tests so that you can run Pytest unit tests immediately.


vrdmr (approved):
LGTM. Thanks for adding the devcontainer configuration.

PR #1049, opened 2022-06-01
Move invocation logging into a thread pool executor

tonybaloney:
This change moves the logging call inside the function invocation callback into a thread pool executor (With a max size of 1).
Effectively allows the logging call to happen concurrently with the function invocation.

In real-terms, with a patched function host, the response times were:

|          | 50 | 66 | 75 | 80 | 90 | 95 | 98 | 99 |
|----------|----|----|----|----|----|----|----|----|
| Original | 15 | 17 | 18 | 20 | 27 | 42 | 58 | 64 |
| Patched  | 14 | 16 | 17 | 19 | 26 | 35 | 47 | 57 |

screenshot 2022-06-01 at 11 37 55

At the 95th percentile an improvement of 7ms per request (18%).

IF the customer is using logging inside their function, we should consider making the thread pool executor available/patched on the logger instance.

### DIFF
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -92,6 +92,10 @@ class Dispatcher(metaclass=DispatcherMeta):
         self._grpc_connected_fut = loop.create_future()
         self._grpc_thread: threading.Thread = threading.Thread(
             name='grpc-thread', target=self.__poll_grpc)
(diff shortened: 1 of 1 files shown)


gavin-aguiar (dismissed):

tonybaloney:
This should be extended to cover the customer logging function code as well. The Python HTTP function demo spends 70% of its time on line 7

image

tonybaloney:
image
With the threaded logger.

PR #1228, opened 2023-05-02
Logging improvements

gavin-aguiar:
## Description

Added a few logging improvements which should improve cold starts. 

Fixes #

---

### PR information


pdthummar (approved):

azure_functions_worker/utils/dependency.py
```diff
@@ -210,17 +210,17 @@ def reload_azure_google_namespace_from_worker_deps(cls):
...
         for p in packages_to_reload:
             try:
-                logger.info('Reloading %s module', p)
                 importlib.reload(sys.modules[p])
+                packages_reloaded.append(p)
             except Exception as ex:
                 logger.info('Unable to reload %s: \n%s', p, ex)
-            logger.info('Reloaded %s module', p)
+
+        logger.info(f'Reloaded modules: {",".join(packages_reloaded)}')
```
vrdmr, line 221: We need the `azure.functions` log as well. Right?

vrdmr: Also, rather than f-string, use the %s method as its a little more performant.

azure_functions_worker/dispatcher.py
```diff
@@ -290,6 +290,8 @@ async def _handle__worker_init_request(self, request):
 
         if DependencyManager.is_in_linux_consumption() \
                 and is_envvar_true(PYTHON_LOAD_FUNCTIONS_INIT):
+            logger.info(
+                "PYTHON_LOAD_FUNCTIONS_INIT enabled. Importing azure functions")
```
vrdmr, line 294: On a second thought - shouldn't this info be within DependencyManager? Where it should make a call on loading or not.

azure_functions_worker/utils/dependency.py
gavin-aguiar: We don't need the "Reloading" log. If it's successful, The "Reloaded" log will have both the packages else the "Unable to reload" log will be shown,



> Also, rather than f-string, use the %s method as its a little more performant.

Also, f-strings in Python are generally faster than the older string formatting methods such as using the "%". T f-strings are evaluated at runtime and don't require additional parsing steps

azure_functions_worker/dispatcher.py
gavin-aguiar: DependencyManager does nothing in this operation other than checking if the app is consumption or not. This log should be in the init request since its being imported at init. Also this is just a temp log till we validate that the functionality works and it will be removed once we make it default.

vrdmr (approved):

PR #1335, opened 2023-10-25
fix for cust sys packages path

hallvictoria:
## Description
Changes in worker and constants to fix "Customer packages not in sys path." warning log being hit.

Fixes #

---

### PR information

### DIFF
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -56,4 +56,4 @@ RETRY_POLICY = "retry_policy"
 # Paths
 CUSTOMER_PACKAGES_PATH = os.path.join(pathlib.Path.home(),
                                       pathlib.Path(
-                                          "site/wwwroot/.python_packages"))
+    "site/wwwroot/.python_packages/lib/site-packages"))
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -1,7 +1,5 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
-import os
-import pathlib
 import sys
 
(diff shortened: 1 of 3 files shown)


azure_functions_worker/constants.py
```diff
@@ -56,4 +56,4 @@
 # Paths
 CUSTOMER_PACKAGES_PATH = os.path.join(pathlib.Path.home(),
                                       pathlib.Path(
-                                          "site/wwwroot/.python_packages"))
+    "site/wwwroot/.python_packages/lib/site-packages"))
```
vrdmr, line 59: The indentation - is flake working?

python/prodV4/worker.py
```diff
@@ -1,10 +1,9 @@
 import os
+import pathlib
 import sys
 
-from pathlib import Path
```
vrdmr, line 4: Any specific reason to remove it?

azure_functions_worker/constants.py
vrdmr: Also - adding /lib/site-packages - would it work? we should also explore if we need another check if site/wwwroot/.python_packages/ exists

And why not add /home here as well? to simplify it

hallvictoria: For python 3.7+ customer packages should be at /home/site/wwwroot/.python_packages/lib/site-packages, so this should fix the issue

python/prodV4/worker.py
hallvictoria: Making it more consistent with the other imports in that file, and "import pathlib" and "pathlib.Path" was how the module was used in constants.py before

gavin-aguiar (approved):

PR #1391, opened 2024-01-05
Enabled implicit output for generic bindings

hallvictoria:
## Description

Fixes #

---

### PR information

### DIFF
--- a/azure_functions_worker/bindings/generic.py
+++ b/azure_functions_worker/bindings/generic.py
@@ -6,6 +6,21 @@ from . import datumdef
 from typing import Any, Optional
 
 
+class GenericBindingProperties:
+
+    def __init__(self, *,
+                 bind_name: Optional[str] = None,):
+        self.__bind_name = bind_name
+
+    @property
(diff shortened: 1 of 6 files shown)


azure_functions_worker/bindings/generic.py
```diff
@@ -52,4 +52,4 @@ def decode(cls, data: datumdef.Datum, *, trigger_metadata) -> typing.Any:
 
     @classmethod
     def has_implicit_output(cls) -> bool:
-        return False
+        return True
```
gavin-aguiar, line 55: We don't need to return False for durable?

hallvictoria: If we create a DurableClient converter in the library, it won't be generic anymore so it won't hit this

gavin-aguiar: We would still need to add this check for backward compatibility. If the customer has pinned the library to an older version, this would fail right?

nytian (approved):
Thanks for the PR!

PR #1411, opened 2024-01-29
Enhance test Webhost logs and fix 503s

YunchuWang:
## Description

1. Add health check for test webhost after starting up
2. Refactor webhost log after test
3. Update deprecated github actions in pipelines
4. Add a flag to archive webhost logs ad-hoc after tests all complete in the pipelines.

---

### PR information


tests/utils/testutils.py
```diff
@@ -287,16 +295,18 @@ def _run_test(self, test, *args, **kwargs):
...
-                self.host_stdout_logger.error(
-                    'Captured WebHost stdout from %s :\n%s',
-                    self.host_stdout.name, self.host_out)
             finally:
-                if test_exception is not None:
-                    raise test_exception
+                try:
+                    self.host_stdout.seek(last_pos)
+                    self.host_out = self.host_stdout.read()
+                    if self.host_out is not None:
```
hallvictoria, line 302: Does this check work? The failing 3.11 E2E tests printed "Captured WebHost log generated during test" but no other logs were printed

YunchuWang: updated.

tests/utils/testutils.py
```diff
@@ -776,16 +799,25 @@ def __init__(self, proc, addr):
         self._proc = proc
         self._addr = addr
 
-    def request(self, meth, funcname, *args, **kwargs):
+    def is_healthy(self):
+        r = self.request('GET')
```
gavin-aguiar, line 803: This should be self..request('GET', '', no_prefix=True) right?

tests/utils/testutils.py
```diff
@@ -776,16 +799,25 @@ def __init__(self, proc, addr):
...
+        return 200 <= r.status_code < 300
+
+    def request(self, meth, func_name=None, *args, **kwargs):
         request_method = getattr(requests, meth.lower())
         params = dict(kwargs.pop('params', {}))
         no_prefix = kwargs.pop('no_prefix', False)
         if 'code' not in params:
             params['code'] = 'testFunctionKey'
+        request_uri = None
```
gavin-aguiar, line 812: We don't need to do this if we pass no-prefix to the request method?

tests/utils/testutils.py
```diff
@@ -257,6 +258,14 @@ def setUpClass(cls):
                 _setup_func_app(TESTS_ROOT / script_dir)
                 cls.webhost = start_webhost(script_dir=script_dir,
                                             stdout=cls.host_stdout)
+            if not cls.webhost.is_healthy():
+                cls.host_out = cls.host_stdout.read()
+                if cls.host_out is not None and len(cls.host_out) > 0:
+                    cls.host_stdout_logger.error(
+                        'WebHost is not started correctly. %s :\n%s',
+                        cls.host_stdout.name, cls.host_out)
+                    raise
```
vrdmr, line 267: What is this raise for?

tests/utils/testutils.py
```diff
@@ -267,6 +276,18 @@ def tearDownClass(cls):
         cls.webhost = None
 
         if cls.host_stdout is not None:
+            if is_envvar_true(ARCHIVE_WEBHOST_LOGS):
+                version_info = sys.version_info
+                log_file = (
+                    "logs/"
+                    f"{cls.__name__}_"
```
vrdmr, line 283: De we need to add [`__file__`](https://docs.python.org/3/reference/import.html#file__) option rather than name, to help with the issue I was mentioning? or that is specific to async tests only?

tests/utils/testutils.py
```diff
@@ -267,6 +276,18 @@ def tearDownClass(cls):
...
+                    f"{version_info.major}.{version_info.minor}_webhost.log"
```
vrdmr, line 284: you can remove the major version to save some length if you want.

tests/utils/testutils.py
```diff
@@ -267,6 +276,18 @@ def tearDownClass(cls):
...
+                )
+                with open(log_file, 'w+') as file:
+                    cls.host_stdout.seek(0)
+                    content = cls.host_stdout.read()
```
vrdmr, line 288: We could do a sanity check that there are some logs (`len(content)`)?

tests/utils/testutils.py
```diff
@@ -287,16 +308,18 @@ def _run_test(self, test, *args, **kwargs):
...
-                    'Captured WebHost stdout from %s :\n%s',
-                    self.host_stdout.name, self.host_out)
             finally:
-                if test_exception is not None:
-                    raise test_exception
+                try:
+                    self.host_stdout.seek(last_pos)
+                    self.host_out = self.host_stdout.read()
+                    if self.host_out is not None and len(self.host_out) > 0:
+                        self.host_stdout_logger.error(
```
vrdmr, line 316: Would this also help that the webhost logs file to look at the is also listed?

tests/utils/testutils.py
YunchuWang: from debug console, i dont see teardownclass param has __file__, 
cls.__name__
'TestLogFilteringFunctions'
cls.__file__
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: type object 'TestLogFilteringFunctions' has no attribute '__file__'

for these sync tests, the module name is the filename. i can add cls.__module__.

tests/utils/testutils.py
YunchuWang: probably not, archiving webhost logs happens in teardown of every test capturing complete webhost log from tmpfile. here is more for not archiving case where we want to log webhost log segment happening for failed tests.

gavin-aguiar (approved):

vrdmr (approved):
:shipit:

PR #1423, opened 2024-02-05
feat: SDK bindings changes

hallvictoria:
## Description
- allows new TypedData object of "model_binding_data"
- imports base library & creates deferred bindings registry if needed
- checks defined type to see if should assign ClientConverter or regular Converter
- adds SupportsDeferredBinding flag to each supported binding through raw_bindings

Fixes #

---

### PR information


tests/utils/testutils.py
```diff
@@ -827,6 +827,7 @@ def popen_webhost(*, stdout, stderr, script_root=FUNCS_PATH, port=None):
         testconfig.read(WORKER_CONFIG)
 
     hostexe_args = []
+    os.environ['AzureWebJobsFeatureFlags'] = 'EnableWorkerIndexing'
```
YunchuWang: is this flag still required?

tests/utils/constants.py
```diff
@@ -2,6 +2,51 @@
...
+      <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.ServiceBus"
+        Version="5.13.6" />
+      <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Sql"
+        Version="3.0.534" />
+      <PackageReference
+        Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator"
+        Version="4.0.1" />
+      <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask"
+        Version="2.13.1" />
+        <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Tables"
```
YunchuWang: init indentation

azure_functions_worker/bindings/datumdef.py
```diff
@@ -93,6 +93,8 @@ def from_typed_data(cls, td: protos.TypedData):
             val = td.collection_string
         elif tt == 'collection_sint64':
             val = td.collection_sint64
+        elif tt == 'model_binding_data':
```
YunchuWang, line 96: codecov warning, is this tested?

azure_functions_worker/bindings/meta.py
```diff
@@ -25,11 +29,38 @@ def load_binding_registry() -> None:
     global BINDING_REGISTRY
     BINDING_REGISTRY = func.get_binding_registry()
 
-
-def get_binding(bind_name: str) -> object:
+    # The SDKs only support python 3.9+
+    if sys.version_info.minor > 8:
+        # Check if cx has imported sdk bindings library
+        try:
+            clients = importlib.util.find_spec('azure.functions.extension.base')
```
YunchuWang: nit: constant

azure_functions_worker/bindings/datumdef.py
YunchuWang: can we go through all the codedev warnings?

azure_functions_worker/bindings/meta.py
YunchuWang: in setup we install extension-base if py ver >= 3.8, so this pkg shall be installed >=3.8 all the time. you can remove  clients = importlib.util.find_spec('azure.functions.extension.base') and directly import to use? correct me if i am wrong

azure_functions_worker/bindings/meta.py
```diff
@@ -25,11 +29,38 @@ def load_binding_registry() -> None:
...
+        if clients is not None:
+            import azure.functions.extension.base as clients
+            global SDK_BINDING_REGISTRY
+            SDK_BINDING_REGISTRY = clients.get_binding_registry()
+
+
+def get_binding(bind_name: str, pytype: typing.Optional[type] = None) -> object:
     binding = None
+
     registry = BINDING_REGISTRY
```
YunchuWang: nit: rename the registry to default_binding_registry or sort, and client_registry to sdk_binding_registry

azure_functions_worker/bindings/meta.py
```diff
@@ -93,6 +125,22 @@ def from_incoming_proto(
         raise TypeError(f'Unknown ParameterBindingType: {pb_type}')
 
     try:
+        # if the binding is an sdk type binding
+        if (SDK_BINDING_REGISTRY is not None
```
YunchuWang: nit: extract this logic to a helper function

azure_functions_worker/bindings/meta.py
YunchuWang: actually these two vars can be removed. we can use the uppercased global vars to cause less confusion.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,11 +29,38 @@ def load_binding_registry() -> None:
...
+
     registry = BINDING_REGISTRY
-    if registry is not None:
+    client_registry = SDK_BINDING_REGISTRY
+    # checks first if registry exists (library is imported)
+    # then checks if pytype is a supported type (cx is using sdk type)
+    if (client_registry is not None
+            and client_registry.check_supported_type(pytype)):
+        global deferred_bindings_enabled
+        deferred_bindings_enabled = True
```
YunchuWang: is it possible we just expose a single function from extension_base like deferred_bindings_enable(pytype)? i see this is checked in multiple places and a single checker function can make it simpler.

YunchuWang: is there a better place to do this? get_binding is called multiple times right, deferred_bindings_enabled shall be set once is enough?

azure_functions_worker/bindings/meta.py
YunchuWang: nit: deferred_bindings_enabled can be used here?

azure_functions_worker/loader.py
```diff
@@ -130,6 +130,13 @@ def process_indexed_function(functions_registry: functions.Registry,
         binding_protos = build_binding_protos(indexed_function)
         retry_protos = build_retry_protos(indexed_function)
 
+        # Check if deferred bindings is enabled
+        if bindings.meta.deferred_bindings_enabled:
```
YunchuWang: null check on chained variables?

azure_functions_worker/loader.py
```diff
@@ -130,6 +130,13 @@ def process_indexed_function(functions_registry: functions.Registry,
...
+            raw_bindings = bindings.meta.SDK_BINDING_REGISTRY.get_raw_bindings(
```
YunchuWang: SDK_BINDING_REGISTRY can be null?

azure_functions_worker/bindings/datumdef.py
hallvictoria: Fixed

azure_functions_worker/loader.py
hallvictoria: It shouldn't be if deferred_bindings_enabled is true, but I can add an extra check to default to the old way of generating raw_bindings just in case

tests/utils/testutils.py
hallvictoria: Good catch, removed

azure_functions_worker/bindings/meta.py
hallvictoria: Yeah makes sense, it shouldn't be needed

azure_functions_worker/loader.py
hallvictoria: Fixed

azure_functions_worker/bindings/meta.py
hallvictoria: Good point, removed

tests/utils/constants.py
hallvictoria: good catch, fixed

azure_functions_worker/bindings/meta.py
hallvictoria: Extracted logic into helper functions. Deferred_bindings_enabled can't be used here. We also need to check the binding type before calling the appropriate decode()

.github/Scripts/e2e-tests.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 python -m pytest -q -n auto --dist loadfile --reruns 4 --instafail --cov=./azure_functions_worker --cov-report xml --cov-branch --cov-append tests/endtoend/test_worker_process_count_functions.py tests/endtoend/test_threadpool_thread_count_functions.py
-python -m pytest -q -n auto --dist loadfile --reruns 4 --instafail --cov=./azure_functions_worker --cov-report xml --cov-branch --cov-append --ignore=tests/endtoend/test_worker_process_count_functions.py --ignore=tests/endtoend/test_threadpool_thread_count_functions.py tests/endtoend
\ No newline at end of file
+python -m pytest -q -n auto --dist loadfile --reruns 4 --instafail --cov=./azure_functions_worker --cov-report xml --cov-branch --cov-append --ignore=tests/endtoend/test_worker_process_count_functions.py --ignore=tests/endtoend/test_threadpool_thread_count_functions.py --ignore=tests/endtoend/test_deferred_bindings_functions.py tests/endtoend
```
gavin-aguiar: Do you think it would be better to have the deferred_binding tests in a separate folder in tests? In that way you wouldn't need to add this line

.github/Scripts/deferred-bindings-e2e-tests.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+python -m pytest -q -n auto --dist loadfile --reruns 4 --instafail --cov=./azure_functions_worker --cov-report xml --cov-branch --cov-append tests/endtoend/test_deferred_bindings_functions.py
```
gavin-aguiar: You don't need this file. Call it directly in the pipeline

.github/workflows/ci_deferred_bindings_workflow.yml
```diff
@@ -0,0 +1,128 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      archive_webhost_logging:
+        description: "For debugging purposes, archive test webhost logs"
+        required: false
+        default: "false"
+  push:
+    branches: [dev, master, main, release/*]
```
gavin-aguiar: Remove master. We don't have that branch

.github/workflows/ci_deferred_bindings_workflow.yml
```diff
@@ -0,0 +1,128 @@
...
+        python-version: [3.9, "3.10", "3.11"]
+    permissions: read-all
+    steps:
+      - name: Checkout code.
+        uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Set up Dotnet 6.x
```
gavin-aguiar: Do we need both dotnet 6 and 8?

.github/workflows/ci_deferred_bindings_workflow.yml
```diff
@@ -0,0 +1,128 @@
...
+                sleep 1
+              fi
+            done
+          }
+
+          python -m pip install --upgrade pip
+          python -m pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple -U azure-functions --pre
+          python -m pip install -U -e .[dev]
+          # python -m pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple --pre .[deferred-bindings]
+          # Installing blob from test pypi right now. Later, can install from .[deferred-bindings]
```
gavin-aguiar: We can remove test.pypi reference now since we pushed it to pypi now

azure_functions_worker/bindings/meta.py
```diff
@@ -13,6 +15,9 @@
 PB_TYPE_DATA = 'data'
 PB_TYPE_RPC_SHARED_MEMORY = 'rpc_shared_memory'
 BINDING_REGISTRY = None
+SDK_BINDING_REGISTRY = None
+deferred_bindings_enabled = False
```
gavin-aguiar: Upper case like the rest

azure_functions_worker/bindings/meta.py
```diff
@@ -25,12 +30,30 @@ def load_binding_registry() -> None:
...
-    binding = None
-    registry = BINDING_REGISTRY
-    if registry is not None:
-        binding = registry.get(bind_name)
+    # The SDKs only support python 3.9+
+    if sys.version_info.minor >= BASE_EXT_SUPPORTED_PY_VERSION:
+        # Check if cx has imported sdk bindings library
+        try:
+            clients = importlib.util.find_spec('azure.functions.extension.base')
+        except ModuleNotFoundError:
```
gavin-aguiar: Use ImportError? It should cover both Import error and MNF error

azure_functions_worker/bindings/meta.py
```diff
@@ -25,12 +30,30 @@ def load_binding_registry() -> None:
...
+
+
+def get_binding(bind_name: str, pytype: typing.Optional[type] = None) -> object:
+    binding = get_deferred_binding(bind_name=bind_name, pytype=pytype)
+
+    # Either cx didn't import library or didn't define sdk type
+    if binding is None and BINDING_REGISTRY is not None:
+        binding = BINDING_REGISTRY.get(bind_name)
     if binding is None:
         binding = generic.GenericBinding
```
gavin-aguiar:
```suggestion
    if binding is None:
        if BINDING_REGISTRY is not None
            binding = BINDING_REGISTRY.get(bind_name)
        else:
            binding = generic.GenericBinding
```

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +216,43 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+
+
+def deferred_bindings_decode(binding: typing.Any,
+                             pb: protos.ParameterBinding, *,
+                             pytype: typing.Optional[type],
+                             datum: typing.Any,
+                             metadata: typing.Any):
+    global SDK_CACHE
+
+    if SDK_CACHE is None:
```
gavin-aguiar: the cache is a dictionary. It will never be none. check if there are any keys in the dictionary

azure_functions_worker/constants.py
```diff
@@ -55,3 +55,6 @@
 
 # Paths
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
+
+# Base extension supported Python version
+BASE_EXT_SUPPORTED_PY_VERSION = 8
```
gavin-aguiar: BASE_EXT_SUPPORTED_PY_MINOR_VERSION

gavin-aguiar: Or just keep it as 3.8 and check the major and minor version

gavin-aguiar (changes requested):

azure_functions_worker/constants.py
hallvictoria: Updated to BASE_EXT_SUPPORTED_PY_MINOR_VERSION

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: Fixed. It still installs from test pypi for testing

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: Removed

.github/Scripts/e2e-tests.sh
hallvictoria: Makes sense, refactored so that these tests have their own folder.

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: Fixed

.github/Scripts/deferred-bindings-e2e-tests.sh
hallvictoria: Fixed

azure_functions_worker/bindings/meta.py
hallvictoria: Fixed

azure_functions_worker/bindings/meta.py
hallvictoria: Fixed

azure_functions_worker/bindings/meta.py
hallvictoria: Edit -- needs more investigation

azure_functions_worker/bindings/meta.py
hallvictoria: I call get() and provide a default value. If the dict is empty or the key doesn't exist, obj will be None as expected

azure_functions_worker/bindings/meta.py
hallvictoria: Because we don't support 3.7 for the extensions but still do in the worker, I don't think adding a helper method in the base extension would improve the check much. We'd have to do a check to see if the base extension has been installed or on the python version e.g:
```
if base_ext is not None
   and base_ext.check_deferred_bindings_enabled():
        ....
```

Good point on the flag. I changed it so that the flag is now set only once, looking through to see if there's another place we can set it.

azure_functions_worker/bindings/meta.py
hallvictoria: There isn't a clean way to improve this check. For generic bindings, `binding = BINDING_REGISTRY.get(bind_name)` will return None. We still need to check if binding is None after this call, otherwise the binding won't be set to `generic.GenericBinding` and the function will return None.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,12 +29,25 @@ def load_binding_registry() -> None:
...
+            # azure.functions.extension isn't loaded in
+            pass
+
+
+def get_binding(bind_name: str, pytype: typing.Optional[type] = None) -> object:
+    binding = get_deferred_binding(bind_name=bind_name, pytype=pytype)
+
+    # Either cx didn't import library or didn't define sdk type
+    if binding is None and BINDING_REGISTRY is not None:
+        binding = BINDING_REGISTRY.get(bind_name)
```
hallvictoria, line 76: todo: change order

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +210,41 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+
+def deferred_bindings_decode(binding: typing.Any,
+                             pb: protos.ParameterBinding, *,
+                             pytype: typing.Optional[type],
+                             datum: typing.Any,
+                             metadata: typing.Any):
+    global SDK_CACHE
+    # Check is the object is already in the cache
+    # If dict is empty or key doesn't exist, obj is None
+    obj = SDK_CACHE.get((pb.name, pytype, datum.value.content), None)
```
hallvictoria: todo: rename from obj. explain what should be returned. rename everything sdk into deferred_bindings.

azure_functions_worker/loader.py
```diff
@@ -239,3 +242,17 @@ def index_function_app(function_path: str):
                          f"{script_file_name}.")
 
     return app.get_functions()
+
+
+def get_fx_raw_bindings(indexed_function, function_info):
+    # Check if deferred bindings is enabled
+    if (bindings.meta is not None
+            and bindings.meta.DEFERRED_BINDINGS_ENABLED
```
hallvictoria: todo: remove bindings.meta none check. move sdk_registry check to get_binding & only do once

azure_functions_worker/bindings/meta.py
hallvictoria: The ordering shouldn't be changed. `bind_name` will be 'blob' for both InputStream and BlobClient types. That means that `BINDING_REGISTRY.get('blob')` will always return a BlobConverter, regardless of the defined type. 

When checking if the binding is a deferred binding first, only one `X_REGISTRY.get()` call will be made. If the check is reversed, a blob trigger with a defined type of BlobClient will call both `BINDING_REGISTRY.get()` and `DEFERRED_BINDINGS_REGISTRY.get()`.

azure_functions_worker/bindings/meta.py
hallvictoria: In order to set the flag, we have to check the type declaration of every binding. If at least one type is declared as an sdk type, then the flag needs to be True. I moved setting it out from get_binding() to validate_function_params(). validate_function_params() is only called once and already does checks on the type declaration. Now, flag will only be set once for function apps using deferred bindings.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,16 +29,34 @@ def load_binding_registry() -> None:
...
+        # Check if binding is deferred binding
+        binding = get_deferred_binding(bind_name=bind_name, pytype=pytype)
+        # Binding is not deferred binding type
+        if binding is None:
+            binding = BINDING_REGISTRY.get(bind_name)
+        # Binding is generic
+        if binding is None:
+            binding = generic.GenericBinding
+        return binding
+    except AttributeError:
```
gavin-aguiar: Add this check in init request

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,63 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+        binding = None
+
+        # Checks if pytype is a supported sdk type
+        if DEFERRED_BINDINGS_REGISTRY.check_supported_type(pytype):
+            # Returns deferred binding converter
+            binding = DEFERRED_BINDINGS_REGISTRY.get(bind_name)
+
+        # This will return None if not a supported type
+        return binding
+    except AttributeError:
```
gavin-aguiar: Remove the try/except and check if DEFERRED_BINDINGS_REGISTRY is none

gavin-aguiar (dismissed):

.github/workflows/ci_deferred_bindings_workflow.yml
```diff
@@ -0,0 +1,123 @@
...
+    - cron: "0 8 * * 1,2,3,4"
+
+jobs:
+  build:
+    name: "Python Deferred Bindings CI Run"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.9, "3.10", "3.11"]
```
vrdmr: Can 3.12 also work?

.github/ISSUE_TEMPLATE/deferred_bindings_feature_request.yml
```diff
@@ -0,0 +1,29 @@
+name: Python Worker Deferred Bindings Feature Request
+description: File a Deferred Bindings Feature report
+title: "Request a feature"
+labels: ["python", "feature"]
```
vrdmr: Add a defbinfings (better naming) label here and bug - for easier identification.

.github/ISSUE_TEMPLATE/deferred_bindings_feature_request.yml
```diff
@@ -0,0 +1,29 @@
...
+      description: A clear and concise description of what you expected to happen.
+      placeholder: What should have occurred?
+
+  - type: textarea
+    id: code-snippet
+    attributes:
+      label: Relevant sample code snipped
+      description: Please copy and paste any relevant code snippet of how you want the feature to be used. (This will be automatically formatted into code, so no need for backticks)
+      render: shell
+
```
vrdmr, line 31: Ask if it is on an existing binding or for a new binding in the future.

.github/ISSUE_TEMPLATE/deferred_bindings_bug_report.yml
```diff
@@ -0,0 +1,64 @@
...
+      description: Please provide detailed step-by-step instructions on how to reproduce the bug.
+      placeholder: |
+        1. Go to the [specific page or section] in the application.
+        2. Click on [specific button or link].
+        3. Scroll down to [specific location].
+        4. Observe [describe what you see, e.g., an error message or unexpected behavior].
+        5. Include any additional steps or details that may be relevant.
+
+  - type: textarea
+    id: code-snipped
```
vrdmr:
```suggestion
    id: code-snippet
```

.github/ISSUE_TEMPLATE/deferred_bindings_bug_report.yml
```diff
@@ -0,0 +1,64 @@
...
+
+  - type: dropdown
+    id: environment
+    attributes:
+      label: Where are you facing this problem?
+      default: 0
+      options:
+        - Local - Core Tools
+        - Production Environment (explain below)
+
```
vrdmr, line 66: Ask for requirements.txt as separate entry.

.github/workflows/ci_deferred_bindings_workflow.yml
```diff
@@ -0,0 +1,123 @@
...
+        required: false
+        default: "false"
+  push:
+    branches: [dev, main, release/*]
+  pull_request:
+    branches: [dev, main, release/*]
+  schedule:
+    # Monday to Thursday 1 AM PDT build
+    # * is a special character in YAML so you have to quote this string
+    - cron: "0 8 * * 1,2,3,4"
```
vrdmr: Lets do it everyday.

.github/workflows/ci_deferred_bindings_workflow.yml
```diff
@@ -0,0 +1,123 @@
...
+                return 1
+              else
+                echo "Attempt $attempt_num failed! Trying again in $attempt_num seconds..."
+                sleep 1
+              fi
+            done
+          }
+
+          python -m pip install --upgrade pip
+          python -m pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple -U azure-functions --pre
```
vrdmr: As we repease these steps across all the pipelines - this should become a script that runs everytime and not duplicated. Please create a task for this.

azure_functions_worker/bindings/meta.py
```diff
@@ -4,6 +4,7 @@
 import typing
 
 from .. import protos
+from ..constants import BASE_EXT_SUPPORTED_PY_MINOR_VERSION
```
vrdmr: If the constant is only being used in this module - refer it here only, no need to add import for this.

vrdmr: unless this is going to be overridable.

azure_functions_worker/bindings/meta.py
```diff
@@ -13,6 +14,9 @@
 PB_TYPE_DATA = 'data'
 PB_TYPE_RPC_SHARED_MEMORY = 'rpc_shared_memory'
 BINDING_REGISTRY = None
+DEFERRED_BINDINGS_REGISTRY = None
+DEFERRED_BINDINGS_ENABLED = False
+DEFERRED_BINDINGS_CACHE = {}
```
vrdmr: It isn't a constant - so shouldn't be caps.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +29,33 @@ def load_binding_registry() -> None:
...
-
-def get_binding(bind_name: str) -> object:
-    binding = None
-    registry = BINDING_REGISTRY
-    if registry is not None:
-        binding = registry.get(bind_name)
+    if BINDING_REGISTRY is None:
+        # If the BINDING_REGISTRY is None, azure-functions hasn't been
+        # loaded in properly.
+        raise AttributeError("BINDING_REGISTRY is None.")
```
vrdmr: This should add more info - what caused it to be none and etc. Site packages etc.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +29,33 @@ def load_binding_registry() -> None:
...
+    if sys.version_info.minor >= BASE_EXT_SUPPORTED_PY_MINOR_VERSION:
+        # Import the base extension
+        try:
+            import azure.functions.extension.base as clients
+            global DEFERRED_BINDINGS_REGISTRY
+            DEFERRED_BINDINGS_REGISTRY = clients.get_binding_registry()
+        except ImportError:
+            # This means that the customer hasn't imported the library.
+            # This isn't an error.
+            pass
```
vrdmr: Should there be log.debug for now - for some debugging purposes

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +29,33 @@ def load_binding_registry() -> None:
...
+
+
+def get_binding(bind_name: str, pytype: typing.Optional[type] = None) -> object:
+    # Check if binding is deferred binding
+    binding = get_deferred_binding(bind_name=bind_name, pytype=pytype)
+    # Binding is not a deferred binding type
```
vrdmr: no need for this comment. make the variable name "deferred_binding"

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +29,33 @@ def load_binding_registry() -> None:
...
+            DEFERRED_BINDINGS_REGISTRY = clients.get_binding_registry()
+        except ImportError:
+            # This means that the customer hasn't imported the library.
+            # This isn't an error.
+            pass
+
+
+def get_binding(bind_name: str, pytype: typing.Optional[type] = None) -> object:
+    # Check if binding is deferred binding
+    binding = get_deferred_binding(bind_name=bind_name, pytype=pytype)
```
vrdmr: Why are we doing this check first rather than the more popular case of none?

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,57 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
         return protos.ParameterBinding(
             name=out_name,
             data=rpc_val)
+
+
+def get_deferred_binding(bind_name: str,
```
vrdmr: please move this function above get-binding (or closer to it) for easy understanding.

azure_functions_worker/bindings/meta.py
```diff
@@ -93,6 +116,14 @@ def from_incoming_proto(
         raise TypeError(f'Unknown ParameterBindingType: {pb_type}')
 
     try:
+        # if the binding is an sdk type binding
+        if (DEFERRED_BINDINGS_REGISTRY is not None
```
vrdmr: Can we simplify `DEFERRED_BINDINGS_REGISTRY is not None` check - to be not done repeatedly?

If `get_binding` returns this info - you won't have to do this check again. A way can be to create a class - Binding which has two attributes - binding_type and binding. And `get_binding` should return object of this class and you make the check - or some other idea.

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,57 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+    binding = None
+
+    # Checks if pytype is a supported sdk type
+    if (DEFERRED_BINDINGS_REGISTRY is not None
+            and DEFERRED_BINDINGS_REGISTRY.check_supported_type(pytype)):
+        # Returns deferred binding converter
+        binding = DEFERRED_BINDINGS_REGISTRY.get(bind_name)
+
+    # This will return None if not a supported type
+    return binding
```
vrdmr:
```suggestion
    # This will return None if not a supported type
    return DEFERRED_BINDINGS_REGISTRY.get(bind_name) if (DEFERRED_BINDINGS_REGISTRY is not None
            and DEFERRED_BINDINGS_REGISTRY.check_supported_type(pytype)) else None
```

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,57 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+                             metadata: typing.Any):
+    # This cache holds deferred binding types (ie. BlobClient, ContainerClient)
+    # That have already been created, so that the worker can reuse the
+    # Previously created type without creating a new one. It allows these
+    # Types to be singleton.
+    global DEFERRED_BINDINGS_CACHE
+
+    # Check if the type is already in the cache
+    # If dict is empty or key doesn't exist, deferred_binding_type is None
+    deferred_binding_type = DEFERRED_BINDINGS_CACHE.get((pb.name, pytype,
```
vrdmr: Return them immediately to - rather than using the variable - as the variable isnt used later.

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,57 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+                                                         datum.value.content),
+                                                        None)
+
+    # If the type is in the cache, return it
```
vrdmr: You can reduce the comments - the code is clear enough

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,57 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+
+def set_deferred_bindings_flag(param_anno: type):
+    # If flag hasn't already been set
+    # If DEFERRED_BINDINGS_REGISTRY is not None
+    # If the binding type is a deferred binding type
+    global DEFERRED_BINDINGS_ENABLED
+    if (not DEFERRED_BINDINGS_ENABLED
+            and DEFERRED_BINDINGS_REGISTRY is not None
+            and DEFERRED_BINDINGS_REGISTRY.check_supported_type(param_anno)):
+        DEFERRED_BINDINGS_ENABLED = True
```
vrdmr: newline at the end

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +215,57 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+    # Otherwise, create the specified type and add it to the cache
+    else:
+        deferred_binding_type = binding.decode(datum, trigger_metadata=metadata,
+                                               pytype=pytype)
+        DEFERRED_BINDINGS_CACHE[(pb.name, pytype, datum.value.content)]\
+            = deferred_binding_type
+        return deferred_binding_type
+
+
+def set_deferred_bindings_flag(param_anno: type):
```
vrdmr: What is the purpose of this method? is it to set the flag in raw_bindings or just to tell if the featue is enabled?

azure_functions_worker/loader.py
```diff
@@ -239,3 +242,18 @@ def index_function_app(function_path: str):
...
 
     return app.get_functions()
+
+
+def get_fx_raw_bindings(indexed_function, function_info):
+    # If the flag is True, we know that:
+    # 1. Library is imported
+    # 2. At least one binding is a defined deferred binding type
+    # 3. DEFERRED_BINDINGS_REGISTRY is not None
+    if bindings.meta.DEFERRED_BINDINGS_ENABLED:
```
vrdmr: This should be an explicit method call - `is_deferred_binding_enabled` - also - this isn't a constant - this method captures it well.

azure_functions_worker/loader.py
```diff
@@ -239,3 +242,18 @@ def index_function_app(function_path: str):
...
+        # Reset the flag
+        bindings.meta.DEFERRED_BINDINGS_ENABLED = False
```
vrdmr: I think I'll need more help with this - we are trying to change a flag between functions?

setup.py
```diff
@@ -109,6 +114,9 @@
         "pandas",
         "numpy",
         "pre-commit"
+    ],
+    "deferred-bindings": [
```
vrdmr: Add test - as it only for testing.

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: We only run the E2E tests 4 days a week. It was to save agent running costs I think. Is that not a concern for these tests?

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: Edit -- not yet. Need the "adding py3.12 to pipelines" PR to go in first. 3.12 requires higher protobuf and grpcio versions

.github/ISSUE_TEMPLATE/deferred_bindings_feature_request.yml
hallvictoria: Added

.github/ISSUE_TEMPLATE/deferred_bindings_bug_report.yml
hallvictoria: Fixed

.github/ISSUE_TEMPLATE/deferred_bindings_bug_report.yml
hallvictoria: Added

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: Moved steps into separate script

azure_functions_worker/bindings/meta.py
hallvictoria: It's also used in setup.py. But I can add it with the protobuf, etc packages that are required for installation for 3.8+.

azure_functions_worker/bindings/meta.py
hallvictoria: I don't think we need it. It's not an error if the customer hasn't imported the library

azure_functions_worker/bindings/meta.py
hallvictoria: Moved to load_binding_registry, which is called in init and env reload

azure_functions_worker/bindings/meta.py
hallvictoria: Same with BINDING_REGISTRY -- updated all 4 variables

azure_functions_worker/bindings/meta.py
hallvictoria: Added in sys.path, sys.modules, and if python-packages path exists

azure_functions_worker/bindings/meta.py
vrdmr: Okay

Do we keep account of which app is using this feature and which isnt?

azure_functions_worker/loader.py
```diff
@@ -239,3 +242,18 @@ def index_function_app(function_path: str):
                          f"{script_file_name}.")
 
     return app.get_functions()
+
+
+def get_fx_raw_bindings(indexed_function, function_info):
+    # If the flag is True, we know that:
```
vrdmr: Make this pydocs

azure_functions_worker/loader.py
```diff
@@ -239,3 +242,18 @@ def index_function_app(function_path: str):
...
+    # 1. Library is imported
+    # 2. At least one binding is a defined deferred binding type
+    # 3. deferred_binding_registry is not None
+    if bindings.meta.deferred_bindings_enabled:
+        # Reset the flag
+        bindings.meta.deferred_bindings_enabled = False
```
vrdmr: I need more info on this specific action.

azure_functions_worker/bindings/meta.py
YunchuWang: image
worker installs base package for 3.8+ in setup.py regardless the feature is used or not. it shall not raise import error here otherwise a serious issue with worker build, customers whether using feature or not shall expect functions not running. so i think we should fail fast here @hallvictoria  @vrdmr

azure_functions_worker/functions.py
```diff
@@ -137,15 +139,19 @@ def validate_function_params(params: dict, bound_params: dict,
 
         input_types: typing.Dict[str, ParamTypeInfo] = {}
         output_types: typing.Dict[str, ParamTypeInfo] = {}
+        deferred_bindings_enabled = False
```
hallvictoria: move above complex types

azure_functions_worker/dispatcher.py
```diff
@@ -678,16 +678,23 @@ def index_functions(self, function_path: str):
...
+                                          deferred_binding_info))
+
                 function_log = "Function Name: {}, Function Binding: {}" \
-                    .format(func.get_function_name(),
-                            [(binding.type, binding.name) for binding in
-                             func.get_bindings()])
+                    .format(func.get_function_name(), bindings_info)
                 indexed_function_logs.append(function_log)
 
             logger.info(
```
hallvictoria, line 704: Add log to see if db enabled

azure_functions_worker/functions.py
```diff
@@ -212,6 +221,10 @@ def validate_function_params(params: dict, bound_params: dict,
                 param_bind_type = 'generic'
             else:
                 param_bind_type = binding.type
+                if (deferred_bindings_enabled
+                        and bindings_utils.meta.DEFERRED_BINDING_REGISTRY.check_supported_type(
+                            param_py_type)):
+                    param_bind_type += "-db"
```
hallvictoria: Add property as part of ParamTypeInfo. Don't change bind_name

.github/workflows/ci_deferred_bindings_workflow.yml
hallvictoria: Different PR

tests/utils/testutils.py
```diff
@@ -57,6 +57,7 @@
 E2E_TESTS_ROOT = TESTS_ROOT / E2E_TESTS_FOLDER
 UNIT_TESTS_FOLDER = pathlib.Path('unittests')
 UNIT_TESTS_ROOT = TESTS_ROOT / UNIT_TESTS_FOLDER
+EXTENSION_TESTS_FOLDER = pathlib.Path('extension_tests')
```
vrdmr, line 60: Is there a special need for `extension_tests` folder and not add within E2E tests?

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +34,43 @@ def load_binding_registry() -> None:
     global BINDING_REGISTRY
     BINDING_REGISTRY = func.get_binding_registry()
 
-
-def get_binding(bind_name: str) -> object:
+    if BINDING_REGISTRY is None:
+        # If the BINDING_REGISTRY is None, azure-functions hasn't been
+        # loaded in properly.
+        raise AttributeError('BINDING_REGISTRY is None. Sys Path:'
```
vrdmr: Lets improve the error - if it shows up to the customer, it should have soem more meaning.

You can mention - BINDING_REGISTRY is none, and is because the worker couldn't find the `azure-function` library.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +34,43 @@ def load_binding_registry() -> None:
...
+                             f'{sys.path}, Sys Module: {sys.modules},'
+                             f'python-packages Path exists: '
```
vrdmr: f-string not needed here. Also, space before python-packages

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +34,43 @@ def load_binding_registry() -> None:
...
+        try:
+            import azurefunctions.extensions.base as clients
+            global DEFERRED_BINDING_REGISTRY
+            DEFERRED_BINDING_REGISTRY = clients.get_binding_registry()
+        except ImportError:
+            logger.info('Base extension not found. '
+                        f'Python version: 3.{sys.version_info.minor}, '
+                        f'Sys path: {sys.path}, '
+                        f'Sys Module: {sys.modules}, '
+                        f'python-packages Path exists: '
```
vrdmr: Same - no fstring.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +34,43 @@ def load_binding_registry() -> None:
...
+
+    # The base extension supports python 3.8+
+    if sys.version_info.minor >= BASE_EXT_SUPPORTED_PY_MINOR_VERSION:
+        # Import the base extension
+        try:
+            import azurefunctions.extensions.base as clients
+            global DEFERRED_BINDING_REGISTRY
+            DEFERRED_BINDING_REGISTRY = clients.get_binding_registry()
+        except ImportError:
+            logger.info('Base extension not found. '
```
vrdmr: should it be`info` for the customer? It'll become very verbose log for no reason.

azure_functions_worker/bindings/meta.py
```diff
@@ -25,15 +34,43 @@ def load_binding_registry() -> None:
...
+                        f'Python version: 3.{sys.version_info.minor}, '
+                        f'Sys path: {sys.path}, '
+                        f'Sys Module: {sys.modules}, '
+                        f'python-packages Path exists: '
+                        f'{os.path.exists(CUSTOMER_PACKAGES_PATH)}.')
+
+
+def get_binding(bind_name: str,
```
vrdmr, line 65: We should explore - for future - I do not like that we keep on doing `get_binding` in an inefficient way with each invocation call.

vrdmr (dismissed):

tests/utils/testutils.py
hallvictoria: We'll be adding more tests for each supported deferred binding type / future extension types. I thought it might be cleaner to keep them separate instead of making our E2E tests folder really large

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +231,42 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+        deferred_bindings_cache[(pb.name,
+                                 pytype,
+                                 datum.value.content)] = deferred_binding_type
+        return deferred_binding_type
+
+
+def check_deferred_bindings_enabled(param_anno: type,
+                                    deferred_bindings_enabled: bool) -> (bool,
+                                                                         bool):
+    # The first bool represents if deferred bindings is enabled at a fx level
```
vrdmr: You have mostly all the functions and methods with the "#" coments. Can you please move these to ``` comments for official PyDoc comments

azure_functions_worker/bindings/meta.py
hallvictoria: Added info about library not being found

azure_functions_worker/bindings/meta.py
hallvictoria: Changed to debug

azure_functions_worker/dispatcher.py
```diff
@@ -516,7 +516,9 @@ async def _handle__invocation_request(self, request):
                     trigger_metadata = None
 
                 args[pb.name] = bindings.from_incoming_proto(
-                    pb_type_info.binding_name, pb,
+                    pb_type_info.binding_name,
+                    pb_type_info.deferred_bindings_enabled,
```
vrdmr: Would it be passed all the times?

azure_functions_worker/bindings/meta.py
hallvictoria: Created future task for this.

gavin-aguiar (approved):

azure_functions_worker/bindings/meta.py
```diff
@@ -184,3 +238,61 @@ def to_outgoing_param_binding(binding: str, obj: typing.Any, *,
...
+
+
+def get_deferred_raw_bindings(indexed_function, input_types):
+    """
+    Calls a method from the base extension that generates the raw bindings
+    for a given function. It also returns logs for that function including
+    the defined binding type and if deferred bindings is enabled for that
+    binding.
+    """
+    raw_bindings, bindings_logs = DEFERRED_BINDING_REGISTRY.get_raw_bindings(
```
vrdmr, line 296: Way more improved :)

azure_functions_worker/dispatcher.py
```diff
@@ -678,21 +680,30 @@ def index_functions(self, function_path: str):
...
-                indexed_functions)
+            fx_metadata_results, fx_bindings_logs = (
+                loader.process_indexed_function(
+                    self._functions,
+                    indexed_functions))
 
             indexed_function_logs: List[str] = []
+            bindings_info = []
             for func in indexed_functions:
+                fx_bindings = fx_bindings_logs.get(func)
```
vrdmr: add log - it is only for log and not the actual bindigns.

tests/extension_tests/deferred_bindings_tests/deferred_bindings_functions/deferred_bindings_enabled/function_app.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+import azure.functions as func
+import azurefunctions.extensions.bindings.blob as bindings
```
vrdmr: as blob or funcblob or something - this can become base for the new sampes.

vrdmr (approved):

azure_functions_worker/dispatcher.py
vrdmr: I meant suffix - log

azure_functions_worker/functions.py
```diff
@@ -41,6 +43,7 @@ def __init__(self, function_name: str, msg: str) -> None:
 
 class Registry:
     _functions: typing.MutableMapping[str, FunctionInfo]
+    _deferred_bindings_enabled: bool = False
```
vrdmr, line 46: task to explore it to become a property in future

azure_functions_worker/dispatcher.py
hallvictoria: Yes, it has a default value. from_incoming_proto also has a default value if it's not passed

YunchuWang (approved):

tests/extension_tests/deferred_bindings_tests/test_deferred_bindings.py
```diff
@@ -0,0 +1,122 @@
...
+                                     "\"BlobName\":"
+                                     "\"test-blobclient-trigger.txt\"}")
+        datum = datumdef.Datum(value=sample_mbd, type='model_binding_data')
+
+        obj = meta.deferred_bindings_decode(binding=binding, pb=pb,
+                                            pytype=BlobClient, datum=datum, metadata={})
+
+        self.assertIsNotNone(obj)
+
+    async def test_check_deferred_bindings_enabled(self):
```
vrdmr, line 107: What is the tuple here? Add a pydoc to explain this test. This definitely needs to be solved.

azure_functions_worker/functions.py
hallvictoria: Yes, added a task for it

vrdmr (approved):

tests/extension_tests/deferred_bindings_tests/test_deferred_bindings.py
hallvictoria: It's a tuple representing if deferred bindings is supported at the app level and for that specific binding. The method has pydocs, I'll add to this test as well

azure_functions_worker/dispatcher.py
hallvictoria: Renamed vars to match with others. (indexed_function_logs & indexed_function_bindings_logs), (func & func_binding_logs)

tests/extension_tests/deferred_bindings_tests/deferred_bindings_functions/deferred_bindings_enabled/function_app.py
hallvictoria: Renamed to blob. Good point. It's imported as "blob" in the extension repo samples, so now it's consistent

PR #1446, opened 2024-03-12
Moving function indexing to init request

gavin-aguiar:
## Description
Included in this PR:
- In the dispatcher, indexing is moved from metadata request to init request and env reload request and results are stored and returned in the metadata response
- Indexing in init is behind an flag: ENABLE_INIT_INDEXING
- Unit tests, e2e tests and consumption tests

Fixes #

---

### PR information

### DIFF
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -55,3 +55,6 @@ RETRY_POLICY = "retry_policy"
 
 # Paths
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
+
+# Flag to index functions in handle init request
+INIT_INDEXING = "INIT_INDEXING"
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -57,6 +57,6 @@ RETRY_POLICY = "retry_policy"
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
 
 # Flag to index functions in handle init request
-PYTHON_ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
+PYTHON_ENABLE_INIT_INDEXING = "PYTHON_ENABLE_INIT_INDEXING"
 
 METADATA_PROPERTIES_WORKER_INDEXED = "worker_indexed"
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -57,6 +57,6 @@ RETRY_POLICY = "retry_policy"
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
 
 # Flag to index functions in handle init request
-ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
+PYTHON_ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
 
 METADATA_PROPERTIES_WORKER_INDEXED = "worker_indexed"
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -57,4 +57,4 @@ RETRY_POLICY = "retry_policy"
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
 
 # Flag to index functions in handle init request
-INIT_INDEXING = "INIT_INDEXING"
+ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
--- a/azure_functions_worker/constants.py
+++ b/azure_functions_worker/constants.py
@@ -58,3 +58,5 @@ CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
 
 # Flag to index functions in handle init request
 ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
+
+METADATA_PROPERTIES_WORKER_INDEXED = "worker_indexed"
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -333,7 +333,7 @@ class Dispatcher(metaclass=DispatcherMeta):
             default_value=f'{PYTHON_SCRIPT_FILE_NAME_DEFAULT}')
 
         logger.info(
-            'Received WorkerMetadataRequest from %s, request ID %s, '
+            'Received load metadata request from  from %s, request ID %s, '
             'script_file_name: %s',
             caller_info, self.request_id, script_file_name)
 
@@ -360,7 +360,7 @@ class Dispatcher(metaclass=DispatcherMeta):
 
         if self._function_metadata_exception:
             return protos.StreamingMessage(
-                request_id=self.request_id,
+                request_id=request.request_id,
                 function_metadata_response=protos.FunctionMetadataResponse(
                     result=protos.StatusResult(
                         status=protos.StatusResult.Failure,
@@ -384,15 +384,17 @@ class Dispatcher(metaclass=DispatcherMeta):
         function_metadata = func_request.metadata
         function_name = function_metadata.name
         function_app_directory = function_metadata.directory
+        function_already_indexed = self._functions.get_function(function_id)
 
         logger.info(
             'Received WorkerLoadRequest, request ID %s, function_id: %s,'
-            'function_name: %s,',
-            self.request_id, function_id, function_name)
+            'function_name: %s, function_already_indexed: %s',
+            self.request_id, function_id, function_name,
+            function_already_indexed)
 
         programming_model = "V2"
         try:
-            if not self._functions.get_function(function_id):
+            if not function_already_indexed:
 
                 if function_metadata.properties.get(
                         METADATA_PROPERTIES_WORKER_INDEXED, False):
@@ -626,7 +628,7 @@ class Dispatcher(metaclass=DispatcherMeta):
             if is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
                 self.load_function_metadata(
                     directory,
-                    caller_info=sys._getframe().f_code.co_name)
+                    caller_info="environment_reload_request")
 
             # Change function app directory
             if getattr(func_env_reload_request,
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -30,7 +30,7 @@ from .constants import (PYTHON_ROLLBACK_CWD_PATH,
                         PYTHON_ENABLE_DEBUG_LOGGING,
                         PYTHON_SCRIPT_FILE_NAME,
                         PYTHON_SCRIPT_FILE_NAME_DEFAULT,
-                        PYTHON_LANGUAGE_RUNTIME, ENABLE_INIT_INDEXING,
+                        PYTHON_LANGUAGE_RUNTIME, PYTHON_ENABLE_INIT_INDEXING,
                         METADATA_PROPERTIES_WORKER_INDEXED)
 from .extension import ExtensionManager
 from .logging import disable_console_logging, enable_console_logging
@@ -301,7 +301,7 @@ class Dispatcher(metaclass=DispatcherMeta):
         # dictionary which will be later used in the invocation request
         bindings.load_binding_registry()
 
-        if is_envvar_true(ENABLE_INIT_INDEXING):
+        if is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
             self.load_function_metadata(
                 worker_init_request.function_app_directory,
                 caller_info="worker_init_request")
@@ -353,7 +353,7 @@ class Dispatcher(metaclass=DispatcherMeta):
         metadata_request = request.functions_metadata_request
         function_app_directory = metadata_request.function_app_directory
 
-        if not is_envvar_true(ENABLE_INIT_INDEXING):
+        if not is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
             self.load_function_metadata(
                 function_app_directory,
                 caller_info="functions_metadata_request")
@@ -401,7 +401,7 @@ class Dispatcher(metaclass=DispatcherMeta):
                     # calling the metadata request. In this case we index the
                     # function and update the workers registry
 
-                    if not is_envvar_true(ENABLE_INIT_INDEXING):
+                    if not is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
                         self.load_function_metadata(
                             function_app_directory,
                             caller_info="functions_load_request")
@@ -623,7 +623,7 @@ class Dispatcher(metaclass=DispatcherMeta):
             # reload_customer_libraries call clears the registry
             bindings.load_binding_registry()
 
-            if is_envvar_true(ENABLE_INIT_INDEXING):
+            if is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
                 self.load_function_metadata(
                     directory,
                     caller_info=sys._getframe().f_code.co_name)
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -324,8 +324,8 @@ class Dispatcher(metaclass=DispatcherMeta):
     def get_function_metadata(self, directory, caller_info):
         """
         This method is called to index the functions in the function app directory
-        and save the results in function_metadata_result or 
-        function_metadata_exception in case of an exception. 
+        and save the results in function_metadata_result or
+        function_metadata_exception in case of an exception.
         """
         script_file_name = get_app_setting(
             setting=PYTHON_SCRIPT_FILE_NAME,
--- a/azure_functions_worker/dispatcher.py
+++ b/azure_functions_worker/dispatcher.py
@@ -30,7 +30,7 @@ from .constants import (PYTHON_ROLLBACK_CWD_PATH,
                         PYTHON_ENABLE_DEBUG_LOGGING,
                         PYTHON_SCRIPT_FILE_NAME,
                         PYTHON_SCRIPT_FILE_NAME_DEFAULT,
-                        PYTHON_LANGUAGE_RUNTIME)
+                        PYTHON_LANGUAGE_RUNTIME, INIT_INDEXING)
 from .extension import ExtensionManager
 from .logging import disable_console_logging, enable_console_logging
 from .logging import (logger, error_logger, is_system_log_category,
@@ -72,8 +72,11 @@ class Dispatcher(metaclass=DispatcherMeta):
         self._function_data_cache_enabled = False
         self._functions = functions.Registry()
         self._shmem_mgr = SharedMemoryManager()
-
         self._old_task_factory = None
+
+        # Used to store metadata returns
+        self.function_metadata_result = None
+        self.function_metadata_exception = None
 
         # We allow the customer to change synchronous thread pool max worker
         # count by setting the PYTHON_THREADPOOL_THREAD_COUNT app setting.
@@ -297,6 +300,9 @@ class Dispatcher(metaclass=DispatcherMeta):
         # dictionary which will be later used in the invocation request
         bindings.load_binding_registry()
 
+        if is_envvar_true(INIT_INDEXING):
+            self.get_function_metadata(worker_init_request)
+
         return protos.StreamingMessage(
             request_id=self.request_id,
             worker_init_response=protos.WorkerInitResponse(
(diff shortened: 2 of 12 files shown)


azure_functions_worker/loader.py
```diff
@@ -218,6 +218,7 @@ def load_function(name: str, directory: str, script_file: str,
                f'{os.path.exists(CUSTOMER_PACKAGES_PATH)}')
 def index_function_app(function_path: str):
     module_name = pathlib.Path(function_path).stem
+    logger.info(f'Loading module {module_name}')
```
hallvictoria: Is this log needed?

azure_functions_worker/constants.py
```diff
@@ -55,3 +55,6 @@
 
 # Paths
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
+
+# Flag to index functions in handle init request
+ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
```
YunchuWang: confirming my understanding, by default is this set to false? if we verify this is working once released, we will remove this to keep customer transparent as this is more related to worker's internal mechanism?

YunchuWang: if this is a public app setting, prefix with PYTHON_?

azure_functions_worker/dispatcher.py
```diff
@@ -72,9 +72,12 @@ def __init__(self, loop: BaseEventLoop, host: str, port: int,
         self._function_data_cache_enabled = False
         self._functions = functions.Registry()
         self._shmem_mgr = SharedMemoryManager()
-
         self._old_task_factory = None
 
+        # Used to store metadata returns
+        self.function_metadata_result = None
```
YunchuWang: nit: _function_metadata_result and same for below

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
+        """
         script_file_name = get_app_setting(
             setting=PYTHON_SCRIPT_FILE_NAME,
             default_value=f'{PYTHON_SCRIPT_FILE_NAME_DEFAULT}')
-        function_path = os.path.join(directory, script_file_name)
 
         logger.info(
-            'Received WorkerMetadataRequest, request ID %s, function_path: %s',
-            self.request_id, function_path)
+            'Received WorkerMetadataRequest from %s, request ID %s, '
```
YunchuWang: nit; this is not always called from workermetadatarequest

azure_functions_worker/dispatcher.py
```diff
@@ -297,6 +301,11 @@ async def _handle__worker_init_request(self, request):
         # dictionary which will be later used in the invocation request
         bindings.load_binding_registry()
 
+        if is_envvar_true(ENABLE_INIT_INDEXING):
+            self.get_function_metadata(
+                directory,
+                caller_info=sys._getframe().f_code.co_name)
```
YunchuWang: any perf overhead doing introspection here? if just for logging purpose, i would suggest hardcoding the caller  function name

azure_functions_worker/dispatcher.py
```diff
@@ -273,6 +276,7 @@ async def _handle__worker_init_request(self, request):
                     )
 
         worker_init_request = request.worker_init_request
+        directory = worker_init_request.function_app_directory
```
YunchuWang: nit: change to func_app_directory for more details?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
             request_id=request.request_id,
             worker_status_response=protos.WorkerStatusResponse())
 
-    async def _handle__functions_metadata_request(self, request):
-        metadata_request = request.functions_metadata_request
-        directory = metadata_request.function_app_directory
+    def get_function_metadata(self, directory, caller_info):
```
YunchuWang: func_app_directory?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
         logger.info(
-            'Received WorkerMetadataRequest, request ID %s, function_path: %s',
-            self.request_id, function_path)
+            'Received WorkerMetadataRequest from %s, request ID %s, '
+            'script_file_name: %s',
+            caller_info, self.request_id, script_file_name)
 
         try:
             validate_script_file_name(script_file_name)
+            function_path = os.path.join(directory, script_file_name)
```
YunchuWang: nit: change to functions_path or func_app_path?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
+            self.function_metadata_exception = self._serialize_exception(ex)
 
+    async def _handle__functions_metadata_request(self, request):
+        metadata_request = request.functions_metadata_request
+        directory = metadata_request.function_app_directory
+
+        if not is_envvar_true(ENABLE_INIT_INDEXING):
+            self.get_function_metadata(
+                directory,
+                caller_info=sys._getframe().f_code.co_name)
```
YunchuWang: same as above

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
-                        use_default_metadata_indexing=True,
-                        result=protos.StatusResult(
-                            status=protos.StatusResult.Success)))
+            self.function_metadata_result = (
+                self.index_functions(function_path)) \
+                if os.path.exists(function_path) else None
 
-            fx_metadata_results = self.index_functions(function_path)
+        except Exception as ex:
+            self.function_metadata_exception = self._serialize_exception(ex)
```
YunchuWang: if worker failed after exception is assigned and before handle_metadata_request is invoked. this error is swallowed. i think we shall raise this in init response because if indexing fails here, it fails whole worker anyways. let me know how u think

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
+                directory,
+                caller_info=sys._getframe().f_code.co_name)
+
+        if self.function_metadata_exception:
             return protos.StreamingMessage(
-                request_id=request.request_id,
+                request_id=self.request_id,
                 function_metadata_response=protos.FunctionMetadataResponse(
-                    function_metadata_results=fx_metadata_results,
                     result=protos.StatusResult(
```
YunchuWang, line 378: we dont need to set use_default_metadata_indexing here?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
-                    default_value=f'{PYTHON_SCRIPT_FILE_NAME_DEFAULT}')
-                validate_script_file_name(script_file_name)
-                function_path = os.path.join(
-                    function_metadata.directory,
-                    script_file_name)
-
-                if function_metadata.properties.get("worker_indexed", False) \
-                        or os.path.exists(function_path):
+
+                if function_metadata.properties.get("worker_indexed", False)\
```
YunchuWang: nit: constant?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,92 @@ async def _handle__worker_status_request(self, request):
...
+                        and not is_envvar_true(ENABLE_INIT_INDEXING):
                     # This is for the second worker and above where the worker
                     # indexing is enabled and load request is called without
                     # calling the metadata request. In this case we index the
                     # function and update the workers registry
-                    _ = self.index_functions(function_path)
+
+                    self.get_function_metadata(directory,
+                                               sys._getframe().f_code.co_name)
```
YunchuWang: ditto

azure_functions_worker/dispatcher.py
```diff
@@ -562,6 +581,7 @@ async def _handle__function_environment_reload_request(self, request):
 
             func_env_reload_request = \
                 request.function_environment_reload_request
+            directory = func_env_reload_request.function_app_directory
 
             # Append function project root to module finding sys.path
             if func_env_reload_request.function_app_directory:
```
YunchuWang, line 612: nit: reuse var above

azure_functions_worker/constants.py
YunchuWang: also this is checked in handle envreload request too, maybe PYTHON_ENABLE_ON_START_INDEXING or PYTHON_ENABLE_FAST_INDEXING?

azure_functions_worker/dispatcher.py
```diff
@@ -587,14 +607,17 @@ async def _handle__function_environment_reload_request(self, request):
...
+            DependencyManager.reload_customer_libraries(directory)
 
             # calling load_binding_registry again since the
             # reload_customer_libraries call clears the registry
             bindings.load_binding_registry()
 
+            if is_envvar_true(ENABLE_INIT_INDEXING):
+                self.get_function_metadata(
+                    directory,
+                    caller_info=sys._getframe().f_code.co_name)
```
YunchuWang: ditto

azure_functions_worker/dispatcher.py
YunchuWang: same case for handle_env_reload

YunchuWang: also for handle_function_load_request for second worker or later workers

azure_functions_worker/loader.py
gavin-aguiar: Nope. Good catch. Added it for debugging. Will remove.

azure_functions_worker/constants.py
gavin-aguiar: Right, this will be set to false by default. Will started enabling by stamp and will be removed later. This is not exposed to the public. Also FAST_INDEXING and START_INDEXING doesn't seem accurate to this use case. ENABLE_INIT_INDEXING is more apt IMO

azure_functions_worker/dispatcher.py
gavin-aguiar: Its not. But it still to get the metadata request. The caller will be logged in caller_info.

azure_functions_worker/dispatcher.py
gavin-aguiar: Its not calling inspect which has a larger overhead but I agree, having a constant will be faster. Will update

azure_functions_worker/dispatcher.py
gavin-aguiar: Really dont have a preference here. Named it directory to keep it consistent in all __handle methods since I think metadata requests has as directory

azure_functions_worker/dispatcher.py
gavin-aguiar: its not the function app path. Its more of a function path for v1 and v2

azure_functions_worker/dispatcher.py
gavin-aguiar: Will fix

azure_functions_worker/dispatcher.py
gavin-aguiar: This exception should be saved and raised only in the metadata response since it is related to indexing. We should not throw these exceptions in init or reload since host treats these requests to complete a specific task.
Saving the error and throwing it in the metadata response will also help us debug at which stage the function app failed with the correct error message.

azure_functions_worker/dispatcher.py
gavin-aguiar: It is using the var above. directory is derived from func_env_reload_request

azure_functions_worker/dispatcher.py
gavin-aguiar: Nope, its an exception

azure_functions_worker/dispatcher.py
```diff
@@ -322,7 +322,7 @@ async def _handle__worker_status_request(self, request):
             request_id=request.request_id,
             worker_status_response=protos.WorkerStatusResponse())
 
-    def get_function_metadata(self, directory, caller_info):
+    def get_function_metadata(self, function_app_directory, caller_info):
```
YunchuWang: nit: change to load_function_metadata? get usually returns something back but here is not case

azure_functions_worker/dispatcher.py
YunchuWang: if it is second worker or later, worker wont receive metadatarequest right? i think we need to at least log the exception when doing the indexing to ensure the exception always gets captured in the log. ex. env_reload succeeded but func indexing actually failed, if for some reason handle metadatarequest is not called, the exception log still shall be recorded for troubleshooting.
image

azure_functions_worker/constants.py
vrdmr: Please do add `Python Functions` prefix

azure_functions_worker/dispatcher.py
gavin-aguiar: Fixed

azure_functions_worker/dispatcher.py
gavin-aguiar: Good point. Fixed this.

YunchuWang (approved):

azure_functions_worker/dispatcher.py
YunchuWang: caller_info=sys._getframe().f_code.co_name

azure_functions_worker/dispatcher.py
vrdmr: I think this doesn't help us. The log line should be clearer - The meaning of "Received WorkerMetadataRequest*" is the grpc request. I don't think this should log should be added at all then (if it impacts cold start or make it debug log only.

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,102 @@ async def _handle__worker_status_request(self, request):
...
-                        use_default_metadata_indexing=True,
-                        result=protos.StatusResult(
-                            status=protos.StatusResult.Success)))
+            self._function_metadata_result = (
+                self.index_functions(function_path)) \
+                if os.path.exists(function_path) else None
 
-            fx_metadata_results = self.index_functions(function_path)
+        except Exception as ex:
+            self._function_metadata_exception = ex
```
vrdmr: This was the issue I had.

Level 1 methods (`_handle__functions_metadata_request`, and `_handle__worker_init_request`) would need to know something that level 2 method (`load_function_metadata`) is doing. Rather, call the `load_function_metadata` method but sorround it with try catch in the level 1 methods itself.

It makes the exceptions explicit and not something that level 1 methods should magically know.

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,102 @@ async def _handle__worker_status_request(self, request):
...
+
+        if not is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
+            self.load_function_metadata(
+                function_app_directory,
+                caller_info="functions_metadata_request")
+
+        if self._function_metadata_exception:
             return protos.StreamingMessage(
-                request_id=request.request_id,
+                request_id=self.request_id,
```
vrdmr: Who's setting the request id? Is the the request id that is being set in init? Why `self`?

azure_functions_worker/constants.py
```diff
@@ -55,3 +55,8 @@
 
 # Paths
 CUSTOMER_PACKAGES_PATH = "/home/site/wwwroot/.python_packages/lib/site-packages"
+
+# Flag to index functions in handle init request
+PYTHON_ENABLE_INIT_INDEXING = "ENABLE_INIT_INDEXING"
```
hallvictoria: If it's a python specific app setting, we can add it to app_setting_manager.py and we can log it

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +322,102 @@ async def _handle__worker_status_request(self, request):
...
         function_id = func_request.function_id
         function_metadata = func_request.metadata
         function_name = function_metadata.name
+        function_app_directory = function_metadata.directory
 
         logger.info(
             'Received WorkerLoadRequest, request ID %s, function_id: %s,'
-            'function_name: %s,', self.request_id, function_id, function_name)
+            'function_name: %s,',
+            self.request_id, function_id, function_name)
```
vrdmr, line 404: Would it help to add the information that it was already found in the `self._functions.get_function(function_id)` call in the log itself? To see the scenario on how much this is useful?

azure_functions_worker/dispatcher.py
```diff
@@ -587,14 +617,17 @@ async def _handle__function_environment_reload_request(self, request):
...
-            DependencyManager.reload_customer_libraries(
-                func_env_reload_request.function_app_directory
-            )
+            DependencyManager.reload_customer_libraries(directory)
 
             # calling load_binding_registry again since the
             # reload_customer_libraries call clears the registry
             bindings.load_binding_registry()
 
+            if is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
```
vrdmr, line 641: We should add some sort of PyDoc to describe what are the actions taking place in the __handle__ methods and add these things in there. Almost like - asking Copilot to summarize this method once and then generating Pydoc

azure_functions_worker/utils/common.py
```diff
@@ -153,4 +153,3 @@ def validate_script_file_name(file_name: str):
     pattern = re.compile(r'^[a-zA-Z0-9_][a-zA-Z0-9_\-]*\.py$')
     if not pattern.match(file_name):
         raise InvalidFileNameError(file_name)
-    return True
```
vrdmr, line 156: Why not return true?

tests/consumption_tests/test_linux_consumption.py
```diff
@@ -242,8 +244,25 @@ def test_opencensus_with_extensions_enabled(self):
...
-                "PYTHON_ENABLE_WORKER_EXTENSIONS": "1",
-                "AzureWebJobsFeatureFlags": "EnableWorkerIndexing"
+                PYTHON_ENABLE_WORKER_EXTENSIONS: "1"
+            })
+            req = Request('GET', f'{ctrl.url}/api/opencensus')
+            resp = ctrl.send_request(req)
+            self.assertEqual(resp.status_code, 200)
+
+    @skipIf(sys.version_info.minor != 10,
+            "This is testing only for python310")
```
vrdmr, line 254: Why 3.10 only? is this due to the image limitations?

tests/endtoend/test_http_functions.py
```diff
@@ -205,6 +205,34 @@ def get_script_dir(cls):
                                             'common_libs_functions_stein'
 
 
+class TestHttpFunctionsWithInitIndexing(TestHttpFunctions):
+
+    @classmethod
+    def setUpClass(cls):
+        os.environ["INIT_INDEXING"] = "1"
```
vrdmr: What's INIT_INDEXING?

azure_functions_worker/dispatcher.py
gavin-aguiar: Actually it does help us finding out from where the request is coming from. This would be good for telemetry. Also it wont add any additional latency than what we already have. This will be logged only once, either from the init/reload request or the metadata request, 
I can change the log statement to read 'Received load metadata request from  .....'

azure_functions_worker/dispatcher.py
gavin-aguiar: I actually didn't change it, just moved it up. It what we currently have https://github.com/Azure/azure-functions-python-worker/blob/dev/azure_functions_worker/dispatcher.py#L351.
But your right, it should be request.requestid

azure_functions_worker/utils/common.py
gavin-aguiar: We don't use the return value anywhere. Its just ignored. We only check if any exception is raised

tests/consumption_tests/test_linux_consumption.py
gavin-aguiar: I'm using an existing function app which is in 3.10. Only adding the indexing app setting and verifying if we get the same result.

azure_functions_worker/constants.py
gavin-aguiar: yup. added

azure_functions_worker/dispatcher.py
gavin-aguiar: Fixed. added original log to metadata request

azure_functions_worker/dispatcher.py
gavin-aguiar: Fixed. Surrounded by  try/catch in the caller

azure_functions_worker/dispatcher.py
gavin-aguiar: Fixed.. Added variable function_already_indexed

azure_functions_worker/dispatcher.py
gavin-aguiar: Agreed. Will take it in the next PR

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +325,117 @@ async def _handle__worker_status_request(self, request):
...
+                        status=protos.StatusResult.Failure,
+                        exception=self._serialize_exception(
+                            self._function_metadata_exception))))
+        else:
+            metadata_result = self._function_metadata_result
 
-        except Exception as ex:
             return protos.StreamingMessage(
-                request_id=self.request_id,
+                request_id=request.request_id,
```
vrdmr, line 386: Why difference between self vs request? In L364, self.request_id is used?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +325,117 @@ async def _handle__worker_status_request(self, request):
...
-                        exception=self._serialize_exception(ex))))
+                        status=protos.StatusResult.Success)))
 
     async def _handle__function_load_request(self, request):
         func_request = request.function_load_request
         function_id = func_request.function_id
         function_metadata = func_request.metadata
         function_name = function_metadata.name
+        function_app_directory = function_metadata.directory
+        function_already_indexed = self._functions.get_function(function_id)
```
vrdmr: Is this needed - this will fail for all the V1 cases. What do we log then?

azure_functions_worker/dispatcher.py
```diff
@@ -313,82 +325,117 @@ async def _handle__worker_status_request(self, request):
...
+
+                if function_metadata.properties.get(
+                        METADATA_PROPERTIES_WORKER_INDEXED, False):
                     # This is for the second worker and above where the worker
                     # indexing is enabled and load request is called without
                     # calling the metadata request. In this case we index the
                     # function and update the workers registry
-                    _ = self.index_functions(function_path)
+
+                    if not is_envvar_true(PYTHON_ENABLE_INIT_INDEXING):
```
vrdmr: Why do this check here? Didn't understand.

If init indexing is done, then it'll aready be indexed.
If init is not done then it'll go throught this call and what is the expectation?

vrdmr (approved):
LGTM

PR #1509, opened 2024-06-11
feat: Support OpenTelemetry Azure monitor distro

lzchen:
## Description

We are planning to support auto instrumentation of [Azure monitor distro](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/monitor/azure-monitor-opentelemetry#azure-monitor-opentelemetry-distro-client-library-for-python) for azure functions. This allows automatic collection of telemetry using OpenTelemetry apis/sdks in users applications in azure functions. The current design is an opt-in model which is controlled by an Appsetting `PYTHON_ENABLE_OPENTELEMETRY`. The worker takes an indirect dependency on the distro [package](https://pypi.org/project/azure-monitor-opentelemetry/) and attempts to load/instrument with it via `configure_azure_monitor()` api provided. The idea is to have this package already preinstalled into the functions image. Until the images are pushed out with this package, you may test this change by manually installing the distro onto your dev environment.

Additionally, any exceptions thrown by the distro are caught and logged accordingly. A telemetry sdk should never cause a user's application to crash.

@gavin-aguiar @vrdmr


lzchen:
@gavin-aguiar 

Since this is a telemetry sdk meant only to be loaded once, I have placed this in `worker_init_request`. Is there anything special that happens during `environment_reload_request` that needs to be done pertaining this in your opinion?

azure_functions_worker/constants.py
```diff
@@ -81,5 +81,15 @@
 # Base extension supported Python minor version
 BASE_EXT_SUPPORTED_PY_MINOR_VERSION = 8
 
+# Appsetting to turn on OpenTelemetry support/features
+# Includes turning on Azure monitor distro to send telemetry to AppInsights
 PYTHON_ENABLE_OPENTELEMETRY = "PYTHON_ENABLE_OPENTELEMETRY"
-PYTHON_ENABLE_OPENTELEMETRY_DEFAULT = True
+PYTHON_ENABLE_OPENTELEMETRY_DEFAULT = False
```
hallvictoria, line 87: Since the feature is opt-in and using a python specific app setting, we should add it to our logs. The app setting can be added in `azure_functions_worker/utils/app_setting_manager.py`

azure_functions_worker/dispatcher.py
```diff
@@ -288,6 +288,46 @@ async def _dispatch_grpc_request(self, request):
...
+            self._otel_libs_available = True
+
+            logger.info("Successfully configured Azure monitor distro.")
+        except ImportError:
+            logger.exception(
+                "Cannot import Azure Monitor distro."
+            )
+            self._otel_libs_available = False
+        except Exception:
+            logger.exception(
```
hallvictoria, line 329: Can we also log the exception that occurred here? Or will it be logged through the sdk?

lzchen: I believe `logger.exception` automatically includes the exception information and traceback if within the context of a `try...except`. https://docs.python.org/3/library/logging.html#logging.Logger.exception

azure_functions_worker/dispatcher.py
```diff
@@ -299,12 +342,11 @@ def update_opentelemetry_status(self):
 
             self._context_api = context_api
             self._trace_context_propagator = TraceContextTextMapPropagator()
-            self._otel_libs_available = True
 
             logger.info("Successfully loaded OpenTelemetry modules. "
                         "OpenTelemetry is now enabled.")
         except ImportError:
-            self._otel_libs_available = False
+            self._azure_monitor_available = False
```
hallvictoria: Do we still need to set `_azure_monitor_available` to false here? It has a default value of false, and it will be set to false in `initialize_azure_monitor` if there's an ImportError

lzchen: Changed to log an exception instead.

azure_functions_worker/dispatcher.py
```diff
@@ -288,6 +291,46 @@ async def _dispatch_grpc_request(self, request):
...
+            os.environ.setdefault(
+                "OTEL_EXPERIMENTAL_RESOURCE_DETECTORS",
+                "azure_functions",
+            )
+
+            configure_azure_monitor(
+                # Connection string can be explicitly specified in Appsetting
+                # If not set, defaults to env var
+                # APPLICATIONINSIGHTS_CONNECTION_STRING
+                connection_string=get_app_setting(
```
gavin-aguiar, line 312: What if the app insight connection string is not set? Should we throw an exception?

azure_functions_worker/dispatcher.py
```diff
@@ -288,6 +291,46 @@ async def _dispatch_grpc_request(self, request):
...
+                    setting=APPLICATIONINSIGHTS_CONNECTION_STRING
+                ),
+                logger_name=get_app_setting(
+                    setting=PYTHON_AZURE_MONITOR_LOGGER_NAME,
+                    default_value=PYTHON_AZURE_MONITOR_LOGGER_NAME_DEFAULT
+                ),
+            )
+            self._azure_monitor_available = True
+
+            logger.info("Successfully configured Azure monitor distro.")
```
gavin-aguiar, line 322: Lets remove this log since we already have a log in the initialize method.

azure_functions_worker/dispatcher.py
lzchen: The distro already throws an exception if this is not configured properly, which will be caught by the exception handler.

azure_functions_worker/dispatcher.py
```diff
@@ -299,12 +342,11 @@ def update_opentelemetry_status(self):
 
             self._context_api = context_api
             self._trace_context_propagator = TraceContextTextMapPropagator()
-            self._otel_libs_available = True
 
             logger.info("Successfully loaded OpenTelemetry modules. "
```
hallvictoria: We can remove this log in `update_opentelemetry_status` in favor of the one added in `initialize_azure_monitor`. It's redundant now

lzchen: Removed.

hallvictoria (approved):

gavin-aguiar (approved):

ashleysommer:
@lzchen 
I've started using this new version of Azure Functions Python Worker locally, and I believe after this addition of Azure Monitor Distro feature there is a regression in the previous behaviour even though this new feature is.

I use OpenTelemetry in my app by calling `context_api.get_context()` to get the current context that is pushed by the Function Worker.
In previous versions, this context was pushed on every invocation if `_otel_libs_available` is True. After this feature addition, the context is pushed on every invocation only if `_azure_monitor_available` is True.

`_azure_monitor_available` is only True under the following condition:
1) `PYTHON_ENABLE_OPENTELEMETRY` var is set _and_
2) OpenTelemetry python modules are installed _and_
3) `azure.monitor.opentelemetry` python module is installed.

In my case, I do not with to use Azure Monitor Otel Distro, so I have:
`PYTHON_ENABLE_OPENTELEMETRY` enabled, _and_ OpenTelemetry  python modules installed, but `azure.monitor.opentelemetry` module _not_ installed.
In pre-v4.31, my application works fine (the worker sets the context using `context_api.attach()`, and I can retrieve it in my app using `context_api.get_context()`. In v4.31+ the worker does not attach the context on each invocation because `_azure_monitor_available` is not True.

lzchen:
@ashleysommer  

We will be rolling out images in which the `azure.monitor.opentelemetry` module will be pre-installed on the image itself so it will always exist (regardless if (1) is set). Thus, effectively, only (1) and (2) would be needed from the users side.

There is an interim stage in which these images will not be pushed to production yet. If you would like to use the preview version of the worker, you can set the context in code [manually](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/monitor/azure-monitor-opentelemetry#trace-correlation) for now as a workaround.

lzchen:
As a followup to the above, we introduced this pr: https://github.com/Azure/azure-functions-python-worker/pull/1621

We have decided to use separate app settings to control context propagation and enabling appinsights explitcily, the latter automatically also supports the former.

PR #1621, opened 2024-12-12
feat: Add env var to control sending telemetry to appinsights in functions worker

lzchen:
Follow up to https://github.com/Azure/azure-functions-python-worker/pull/1509

We have decided to separate out the context propagation and enabling azure monitor functionality. To do this, we will be using a separate env var `PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY`, which will be for strictly sending telemetry to appinsights and as well enabling context propagation. Future configuration of the telemetry SDK will most likely have the word `_APPLICATIONINSIGHTS_` inside it's name.

We treating Opentelemetry context propagation as a separate feature. This will be controlled as an opt-in like before with `PYTHON_ENABLE_OPENTELEMETRY` env var.


ghost:
If both `appinsights_app_setting` and `opentelemetry_app_setting` are set to `"true"`, `configure_azure_monitor` could be called twice, first by user then by Dispatcher.
Could this be a potential bug? In case it could be, shouldn't the settings be exclusive with `opentelemetry_app_setting` having higher priority?

lzchen:
@MatteoMilanese 

> configure_azure_monitor could be called twice, first by user then by Dispatcher.

Good question. We do not safeguard against double instrumentation as this is an unsupported scenario. We advise customers to either turn the appsetting on OR use `configure_azure_monitor()` in their code.

azure_functions_worker/dispatcher.py
```diff
@@ -381,12 +385,24 @@ async def _handle__worker_init_request(self, request):
...
+        opentelemetry_app_setting = get_app_setting(
+            setting=PYTHON_ENABLE_OPENTELEMETRY,
+            default_value=PYTHON_ENABLE_OPENTELEMETRY_DEFAULT,
+        )
+
+        appinsights_app_setting = get_app_setting(
+            setting=PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY,
+            default_value=PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY_DEFAULT,
+        )
+        if appinsights_app_setting and appinsights_app_setting.lower() == "true":
```
hallvictoria: 394 - 398 can be simplified to `if is_envvar_true(PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY):`

azure_functions_worker/dispatcher.py
```diff
@@ -381,12 +385,24 @@ async def _handle__worker_init_request(self, request):
...
             self.initialize_azure_monitor()
 
-            if self._azure_monitor_available:
-                capabilities[constants.WORKER_OPEN_TELEMETRY_ENABLED] = _TRUE
+        if opentelemetry_app_setting and opentelemetry_app_setting.lower() == "true":
```
hallvictoria: Same applies here -- use `is_envvar_true(PYTHON_ENABLE_OPENTELEMETRY)` instead of 389-391

azure_functions_worker/dispatcher.py
```diff
@@ -782,11 +798,15 @@ async def _handle__function_environment_reload_request(self, request):
             if get_app_setting(
                     setting=PYTHON_ENABLE_OPENTELEMETRY,
                     default_value=PYTHON_ENABLE_OPENTELEMETRY_DEFAULT):
+                self._otel_libs_available = True
+            if get_app_setting(
+                    setting=PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY,
+                    default_value=PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY_DEFAULT):
```
hallvictoria: If `PYTHON_ENABLE_OPENTELEMETRY` and `PYTHON_APPLICATIONINSIGHTS_ENABLE_TELEMETRY` are only accepted to be True or False, I feel like using `is_envvar_true` is better than `get_app_setting`

tests/unittests/test_opentelemetry.py
```diff
@@ -103,8 +106,113 @@ def test_init_request_otel_capability_disabled_app_setting(
...
+                         protos.StatusResult.Success)
+
+        # Verify otel_libs_available is set to True
+        self.assertTrue(self.dispatcher._otel_libs_available)
+        # Verify that WorkerOpenTelemetryEnabled capability is set to _TRUE
+        capabilities = init_response.worker_init_response.capabilities
+        self.assertIn("WorkerOpenTelemetryEnabled", capabilities)
+        self.assertEqual(capabilities["WorkerOpenTelemetryEnabled"], "true")
+
+    @patch.dict(os.environ, {})
```
hallvictoria: Should this be `@patch.dict(os.environ, {'PYTHON_ENABLE_OPENTELEMETRY': 'false'})`?

hallvictoria:
The unit tests are failing because of `ModuleNotFoundError: No module named 'opentelemetry'` and `ModuleNotFoundError: No module named 'azure.monitor'`. Could you please add these packages to the worker's dev dependencies in pyproject.toml?

PR #1666, opened 2025-04-11
feat: Adding support for python 3.13

gavin-aguiar:
## Description
**Proxy Worker Changes** (merge to dev)
- Adds 3.13 to the build & test pipelines
   - For python 3.12, the builds include azure_functions_worker ONLY
   - For python 3.13+, the builds include proxy_worker ONLY
   - We reuse the existing build / testing structure from the current worker, but we make some logical changes to test afw or proxy accordingly based on python version
- Proxy_worker: receives & returns gRPC messages from the host
   - Majority of the worker functionality is defined in the library worker. The proxy worker will solely receive messages from the host, forward them to the library worker, and respond back.
   - Both the V1 and V2 library worker are included in the image. The proxy worker decides what to import based on the function path.

Fixes #

---

### PR information


proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+                    sys.version,
+                    VERSION,
+                    self.request_id)
+
+        if DependencyManager.should_load_cx_dependencies():
+            DependencyManager.prioritize_customer_dependencies()
+
+        global _library_worker
+        directory = request.worker_init_request.function_app_directory
+        v2_directory = os.path.join(directory, 'function_app.py')
```
hallvictoria: For supporting a configurable file name, this should change to use the app setting instead of hard-coding 'function_app.py'

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+        if os.path.exists(v2_directory):
+            try:
+                import azure_functions_worker_v2  # NoQA
+                _library_worker = azure_functions_worker_v2
+                logger.debug("V2 worker import succeeded: %s", _library_worker.__file__)
+            except ImportError:
+                logger.warning("Error importing V2 library: %s", traceback.format_exc())
```
hallvictoria: Should these be logger.error?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+        return protos.StreamingMessage(
+            request_id=self.request_id,
+            worker_init_response=init_response)
+
+    async def _handle__function_environment_reload_request(self, request):
+        logger.info('Received FunctionEnvironmentReloadRequest, '
+                    'request ID: %s, '
+                    'To enable debug level logging, please refer to '
+                    'https://aka.ms/python-enable-debug-logging',
+                    self.request_id)
```
hallvictoria, line 437: We logged the app setting state here previously. That log was incorrect, but should we still log the updated app settings somewhere later in this method?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+                    'python version %s, '
+                    'worker version %s, '
+                    'request ID %s. '
+                    'To enable debug level logging, please refer to '
+                    'https://aka.ms/python-enable-debug-logging',
+                    sys.version,
+                    VERSION,
+                    self.request_id)
+
+        if DependencyManager.should_load_cx_dependencies():
```
hallvictoria, line 414: Do we need the if statement? This will always be true, right?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+                    'https://aka.ms/python-enable-debug-logging',
+                    self.request_id)
+
+        func_env_reload_request = \
+            request.function_environment_reload_request
+        directory = func_env_reload_request.function_app_directory
+        DependencyManager.prioritize_customer_dependencies(directory)
+
+        global _library_worker
+        directory = func_env_reload_request.function_app_directory
```
hallvictoria: This line is a duplicate of 419

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+        self._grpc_max_msg_len: int = grpc_max_msg_len
+        self._old_task_factory = None
+
+        self._grpc_resp_queue: queue.Queue = queue.Queue()
+        self._grpc_connected_fut = loop.create_future()
+        self._grpc_thread: threading.Thread = threading.Thread(
+            name='grpc_local-thread', target=self.__poll_grpc)
+
+        # TODO: Need to find a better place for these
+        self._function_data_cache_enabled = False
```
hallvictoria: todo?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+                                     properties={"protos": protos,
+                                                 "host": self._host})
+        init_response = await _library_worker.worker_init_request(init_request)
+
+        return protos.StreamingMessage(
+            request_id=self.request_id,
+            worker_init_response=init_response)
+
+    async def _handle__function_environment_reload_request(self, request):
+        logger.info('Received FunctionEnvironmentReloadRequest, '
```
hallvictoria, line 433: Today this is currently wrapped in a try catch. Should we also do that here?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+        global _library_worker
+        directory = func_env_reload_request.function_app_directory
+        v2_directory = os.path.join(directory, 'function_app.py')
+        if os.path.exists(v2_directory):
+            try:
+                import azure_functions_worker_v2  # NoQA
+                _library_worker = azure_functions_worker_v2
+                logger.debug("V2 worker import succeeded: %s", _library_worker.__file__)
+            except ImportError:
+                logger.warning("Error importing V2 library: %s", traceback.format_exc())
```
hallvictoria: slight nit: in the debug log, we refer to the library worker as "V2 worker", and in the warning log, we say "V2 library". Should we call it the same thing for both logs?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+        # Having log here will reduce the responsiveness of the worker.
+        return protos.StreamingMessage(
+            request_id=request.request_id,
+            worker_status_response=protos.WorkerStatusResponse())
+
+    async def _handle__functions_metadata_request(self, request):
+        logger.info(
+            'Received WorkerMetadataRequest, request ID %s, '
+            'worker id: %s',
+            self.request_id, self.worker_id)
```
hallvictoria, line 470: Do we also want to log the function_path here? Why do we need the worker id logged?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+    async def _handle__function_load_request(self, request):
+        func_request = request.function_load_request
+        function_id = func_request.function_id
+        function_metadata = func_request.metadata
+        function_name = function_metadata.name
+
+        logger.info(
+            'Received WorkerLoadRequest, request ID %s, function_id: %s,'
+            'function_name: %s, worker_id: %s',
+            self.request_id, function_id, function_name, self.worker_id)
```
hallvictoria, line 490: Same q here -- why log the worker ID and remove the function app directory?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+
+    async def _handle__invocation_request(self, request):
+        invoc_request = request.invocation_request
+        invocation_id = invoc_request.invocation_id
+        function_id = invoc_request.function_id
+
+        logger.info(
+            'Received FunctionInvocationRequest, request ID %s, function_id: %s,'
+            'invocation_id: %s, worker_id: %s',
+            self.request_id, function_id, invocation_id, self.worker_id)
```
hallvictoria, line 509: Our current function invocation req log:

```
            function_invocation_logs: List[str] = [
                'Received FunctionInvocationRequest',
                f'request ID: {self.request_id}',
                f'function ID: {function_id}',
                f'function name: {fi.name}',
                f'invocation ID: {invocation_id}',
                f'function type: {"async" if fi.is_async else "sync"}',
                f'timestamp (UTC): {invocation_time}'
            ]
````

File: proxy_worker/dispatcher.py [hunk 12]
```diff
@@ -0,0 +1,509 @@
...
+        function_id = func_request.function_id
+        function_metadata = func_request.metadata
+        function_name = function_metadata.name
+
+        logger.info(
+            'Received WorkerLoadRequest, request ID %s, function_id: %s,'
+            'function_name: %s, worker_id: %s',
+            self.request_id, function_id, function_name, self.worker_id)
+
+        load_request = WorkerRequest(name="FunctionsLoadRequest", request=request)
```
hallvictoria: nit: name should be FunctionLoadRequest (singular function)

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
...
+        invoc_request = request.invocation_request
+        invocation_id = invoc_request.invocation_id
+        function_id = invoc_request.function_id
+
+        logger.info(
+            'Received FunctionInvocationRequest, request ID %s, function_id: %s,'
+            'invocation_id: %s, worker_id: %s',
+            self.request_id, function_id, invocation_id, self.worker_id)
+
+        invocation_request = WorkerRequest(name="WorkerInvRequest", request=request,
```
hallvictoria: nit: `name="FunctionInvocationRequest"`

proxy_worker/start_worker.py
```diff
@@ -0,0 +1,84 @@
...
+    import asyncio
+
+    from . import logging
+    from .logging import error_logger, logger
+
+    args = parse_args()
+    logging.setup(log_level=args.log_level, log_destination=args.log_to)
+
+    logger.info("Args: %s", args)
+    logger.info('Starting Azure Functions Python Worker.')
```
hallvictoria: Should we change this to say proxy worker?

proxy_worker/start_worker.py
```diff
@@ -0,0 +1,84 @@
...
+        error_logger.exception(
+            'unhandled error in functions worker: {0}'.format(
+                ''.join(traceback.format_exception(ex))))
+        raise
+
+
+async def start_async(host, port, worker_id, request_id):
+    from . import dispatcher
+
+    # ToDo: Fix functions_grpc_max_msg_len. Needs to be parsed from args
```
hallvictoria: todo?

proxy_worker/utils/app_settings.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import os
+from typing import Callable, Optional
+
+
+def get_app_setting(
```
hallvictoria: Why is this moved out of common.py? common.py also handles environment variables

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,509 @@
+import asyncio
```
hallvictoria, line 4: Add license header

proxy_worker/__init__.py
hallvictoria, line 1: Add license

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,308 @@
...
+        cls.cx_deps_path = cls._get_cx_deps_path()
+        cls.cx_working_dir = cls._get_cx_working_dir()
+        cls.worker_deps_path = cls._get_worker_deps_path()
+
+    @classmethod
+    def is_in_linux_consumption(cls):
+        return CONTAINER_NAME in os.environ
+
+    @classmethod
+    def should_load_cx_dependencies(cls):
```
hallvictoria, line 68: Do we need this if we always enable dependency isolation?

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,308 @@
...
+        # Try to get the latest customer's dependency path
+        cx_deps_path: str = cls._get_cx_deps_path()
+
+        if not cx_deps_path:
+            cx_deps_path = cls.cx_deps_path
+
+        logger.info(
+            'Applying prioritize_customer_dependencies: '
+            'worker_dependencies_path: %s, customer_dependencies_path: %s, '
+            'working_directory: %s, Linux Consumption: %s, Placeholder: %s, '
```
hallvictoria, line 140: Should we add or change LC to FC in the log?

proxy_worker/version.py
```diff
@@ -0,0 +1 @@
+VERSION = "1.0.0"
```
hallvictoria: 4.36.1

hallvictoria: Add license

python/proxyV4/worker.py
```diff
@@ -0,0 +1,65 @@
+import os
```
hallvictoria, line 4: We don't have the license on prodV4/worker.py, but we should probably add it here

python/proxyV4/worker.py
```diff
@@ -0,0 +1,65 @@
...
+if __name__ == '__main__':
+    validate_python_version()
+    func_worker_dir = str(pathlib.Path(__file__).absolute().parent)
+    env = os.environ
+
+    # Setting up python path for all environments to prioritize
+    # third-party user packages over worker packages in PYTHONPATH
+    user_pkg_paths = determine_user_pkg_paths()
+    joined_pkg_paths = os.pathsep.join(user_pkg_paths)
+    env['PYTHONPATH'] = f'{joined_pkg_paths}:{func_worker_dir}'
```
hallvictoria, line 54: future improvement: we've received a couple of complaints about always overwriting PYTHONPATH. We could look into appending to PYTHONPATH instead of setting it

proxy_worker/dispatcher.py
gavin-aguiar: Updated.

proxy_worker/dispatcher.py
gavin-aguiar: Since these should never be logged, keeping this as a warning. This will show up as a yellow line in core tools if it gets logged. Error will show as a red which might be confusing

proxy_worker/dispatcher.py
gavin-aguiar: I think app settings should only be logged in the library. There are only a couple of app settings for that are proxy specific but they too can be logged in the library

proxy_worker/dispatcher.py
gavin-aguiar: yes we do. This also checks that we don't need to run prioritize dependencies on  placeholder mode.

proxy_worker/dispatcher.py
gavin-aguiar: removed

proxy_worker/dispatcher.py
gavin-aguiar: Removed. This setting is used for shared memory which is no longer supported

proxy_worker/dispatcher.py
gavin-aguiar: The library request should be wrapped in a try catch and only send the proxy a success/failure streaming  message.

proxy_worker/dispatcher.py
gavin-aguiar: Updated

proxy_worker/dispatcher.py
gavin-aguiar: I think we added function_path to check if the function is V1/V2. We have other logs for that now. 
Also we do need worker id for cases where there are more than one worker and we need to check which worker receives the metadata request and which gets the load request. It does gets logged as a separate column in kusto but better to have it in the worker logs too.,

proxy_worker/dispatcher.py
gavin-aguiar: ^

proxy_worker/dispatcher.py
gavin-aguiar: We might need to add another log for sync/async in the library. Can it be a part of that big log?

proxy_worker/dispatcher.py
gavin-aguiar: Updated. Does the library check this ?

proxy_worker/dispatcher.py
gavin-aguiar: Updated

proxy_worker/start_worker.py
gavin-aguiar: Updated

proxy_worker/start_worker.py
gavin-aguiar: Removed. TBH I don't remember why did I add this todo

proxy_worker/utils/app_settings.py
gavin-aguiar: Moved it back to commons and deleted the other file

proxy_worker/dispatcher.py
gavin-aguiar: added

python/proxyV4/worker.py
gavin-aguiar: The worker should set while the function app should append

python/proxyV4/worker.py
gavin-aguiar: added

proxy_worker/version.py
gavin-aguiar: added

proxy_worker/utils/dependency.py
gavin-aguiar: Both are consumption v1 and v2. The worker will not know the dfference between the two

proxy_worker/utils/dependency.py
gavin-aguiar: yes

proxy_worker/__init__.py
gavin-aguiar: added

azure_functions_worker/protos/.gitignore
vrdmr, line 1: Why any change in this?

eng/templates/official/jobs/build-artifacts.yml
```diff
@@ -62,6 +65,9 @@ jobs:
       Python312V4:
         pythonVersion: '3.12'
         workerPath: 'python/prodV4/worker.py'
+      Python313V4:
+        pythonVersion: '3.13'
+        workerPath: 'python/proxyV4/worker.py'
```
vrdmr, line 70: Can the worker Path be a variable at the top?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,518 @@
...
+                    'python version %s, '
+                    'worker version %s, '
+                    'request ID %s. '
+                    'To enable debug level logging, please refer to '
+                    'https://aka.ms/python-enable-debug-logging',
+                    sys.version,
+                    VERSION,
+                    self.request_id)
+
+        if DependencyManager.should_load_cx_dependencies():
```
hallvictoria: We always enable dependency isolation. This check prevents doing so during placeholder mode, as it is unnecessary.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,518 @@
...
+            DependencyManager.prioritize_customer_dependencies()
+
+        global _library_worker
+        directory = request.worker_init_request.function_app_directory
+        v2_directory = os.path.join(directory, get_script_file_name())
+        if os.path.exists(v2_directory):
```
hallvictoria: Both the V1 & V2 library worker are included in the image now. For customers, this prevents V1 apps from requiring any changes. For internal, this prevents complicated logic for reloading and resetting the library worker based on if the app is V1 or V2.

We now import either the V1 or V2 library based on the function directory. If the customer brings `function_app.py` or sets the file name via the app setting, then we import V2. Otherwise, V1 is imported. Logs are included as debug to help potential bug investigations.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,518 @@
...
+                             _library_worker.__file__)
+            except ImportError:
+                logger.debug("azure_functions_worker_v1 library not found: %s",
+                             traceback.format_exc())
+
+        init_request = WorkerRequest(name="WorkerInitRequest",
+                                     request=request,
+                                     properties={"protos": protos,
+                                                 "host": self._host})
+        init_response = await _library_worker.worker_init_request(init_request)
```
hallvictoria: The bulk of the worker functionality is defined in the library worker. The proxy worker simply forwards the necessary information to the library worker and waits for the response.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,518 @@
...
+
+        return protos.StreamingMessage(
+            request_id=self.request_id,
+            function_environment_reload_response=env_reload_response)
+
+    async def _handle__worker_status_request(self, request):
+        # Logging is not necessary in this request since the response is used
+        # for host to judge scale decisions of out-of-proc languages.
+        # Having log here will reduce the responsiveness of the worker.
+        return protos.StreamingMessage(
```
hallvictoria, line 462: This method does not involve the library worker and is handled solely by the proxy worker.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,518 @@
...
+            return True
+
+        max_workers = get_app_setting(setting=PYTHON_THREADPOOL_THREAD_COUNT,
+                                      validator=tp_max_workers_validator)
+
+        # We can box the app setting as int for earlier python versions.
+        return int(max_workers) if max_workers else None
+
+    async def _handle__worker_init_request(self, request):
+        logger.info('Received WorkerInitRequest, '
```
hallvictoria: There are three new info logs added with the proxy / library worker split. In the library worker, we log that the library worker has received an init req, an env reload req, and if the function is sync/async in invoc req. The information that is logged does not change with this split -- the new logs added are intended to help confirm that the library worker is receiving requests appropriately.

proxy_worker/logging.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import logging
+import logging.handlers
+import sys
+from typing import Optional
+
+# Logging Prefixes
+SYSTEM_LOG_PREFIX = "proxy_worker"
```
hallvictoria, line 10: The log prefix changes for proxy_worker, but other than that, there are no notable changes.

proxy_worker/utils/constants.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# App Setting constants
```
hallvictoria, line 4: App settings are majorly handled in the library worker. The constants included here are the only ones that need to be handled at the proxy worker level: enabling debug logging, applying PTPTC, and checking the script file name.

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,308 @@
...
+         2) App is linux consumption but not in placeholder mode.
+         This can happen when the worker restarts for any reason
+         (OOM, timeouts etc) and env reload request is not called.
+        """
+        return not (DependencyManager.is_in_linux_consumption()
+                    and is_envvar_true("WEBSITE_PLACEHOLDER_MODE"))
+
+    @classmethod
+    def use_worker_dependencies(cls):
+        """Switch the sys.path and ensure the worker imports are loaded from
```
hallvictoria, line 81: `use_worker_dependencies` is used only when starting up the worker.

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,308 @@
...
+
+        cls._remove_from_sys_path(cls.cx_deps_path)
+        cls._remove_from_sys_path(cls.cx_working_dir)
+        cls._add_to_sys_path(cls.worker_deps_path, True)
+        logger.info('Start using worker dependencies %s. Sys.path: %s',
+                    cls.worker_deps_path, sys.path)
+
+    @classmethod
+    def prioritize_customer_dependencies(cls, cx_working_dir=None):
+        """Switch the sys.path and ensure the customer's code import are loaded
```
hallvictoria, line 106: Since dependency isolation is now enabled by default, this will be called in init and env reload.

proxy_worker/version.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+VERSION = "4.36.1"
```
hallvictoria, line 4: proxy_worker and azure_functions_worker will both be released in the same nuget package. This means that a worker release will include both a proxy_worker (3.13+) and azure_functions_worker (<=3.12) release, so the version must be consistent between pw and afw. 

To look into: top-level version file?

pyproject.toml
```diff
@@ -27,14 +27,19 @@ classifiers = [
...
-     "grpcio-tools~=1.59.0; python_version >= '3.8'",
+     "grpcio-tools~=1.59.0; python_version >= '3.8' and python_version < '3.13'",
+     "grpcio-tools~=1.70.0; python_version >= '3.13'",
      "grpcio~=1.43.0; python_version == '3.7'",
-     "grpcio~=1.59.0; python_version >= '3.8'",
-     "azurefunctions-extensions-base; python_version >= '3.8'"
+     "grpcio ~=1.59.0; python_version >= '3.8' and python_version < '3.13'",
+     "grpcio~=1.70.0; python_version >= '3.13'",
+     "azurefunctions-extensions-base; python_version >= '3.8'",
+     "test-worker==1.0.0a38; python_version >= '3.13'",
```
hallvictoria: We're not bumping grpc / protobuf for existing python versions, but we do have to update these for python 3.13.

TODO: test-worker and test-worker-v1 will change to the released library worker package.

python/test/worker.py
```diff
@@ -16,4 +15,10 @@ def add_script_root_to_sys_path():
 
 if __name__ == '__main__':
     add_script_root_to_sys_path()
-    main.main()
+    minor_version = sys.version_info[1]
```
hallvictoria, line 18: `python/test/worker.py` is used for running tests / debugging locally. Now depending on python version, either pw or afw will be run.

tests/test_setup.py
```diff
@@ -38,6 +37,7 @@
 WEBHOST_GITHUB_API = "https://api.github.com/repos/Azure/azure-functions-host"
 WEBHOST_GIT_REPO = "https://github.com/Azure/azure-functions-host/archive"
 WEBHOST_TAG_PREFIX = "v4."
+WORKER_DIR = "azure_functions_worker" if sys.version_info.minor < 13 else "proxy_worker"
```
hallvictoria, line 40: This is where we differentiate whether the build should include afw or pw. We were previously hard-coding `azure_functions_worker`, and now we either pull in that or proxy_worker depending on python version.

tests/test_setup.py
```diff
@@ -183,8 +183,26 @@ def gen_grpc():
     # Needed to support absolute imports in files. See
     # https://github.com/protocolbuffers/protobuf/issues/1491
     make_absolute_imports(compiled_files)
+    copy_tree_merge(str(built_protos_dir), str(proto_root_dir))
```
hallvictoria, line 186: `dir_util.copy_tree` is deprecated, but it is used as part of our builds to correcly copy over the built `protos` folder. This change replicates what `dir_util` did previously.

tests/utils/testutils.py
```diff
@@ -50,18 +50,23 @@
...
-)
-from azure_functions_worker.bindings.shared_memory_data_transfer import (
-    SharedMemoryConstants as consts,
-)
-from azure_functions_worker.constants import (
-    FUNCTIONS_WORKER_SHARED_MEMORY_DATA_TRANSFER_ENABLED,
-    UNIX_SHARED_MEMORY_DIRECTORIES,
-)
-from azure_functions_worker.utils.common import get_app_setting, is_envvar_true
+if sys.version_info.minor < 13:
```
hallvictoria, line 53: When running the tests, we want to use imports & methods from `azure_functions_worker` if using <=3.12 and those from `proxy_worker` if using 3.13+. This allows for reusing the existing testing structure for all python versions.

eng/templates/official/jobs/ci-e2e-tests.yml
```diff
@@ -63,6 +63,15 @@ jobs:
           SQL_CONNECTION: $(LinuxSqlConnectionString312)
           EVENTGRID_URI: $(LinuxEventGridTopicUriString312)
           EVENTGRID_CONNECTION: $(LinuxEventGridConnectionKeyString312)
+        Python313:
+          PYTHON_VERSION: '3.13'
```
hallvictoria, line 67: 3.13 is added to the E2E and emulator pipelines right now. E2E tests are re-using 312 resources for now until the appropriate 313 resources are created (cosmosdb, sql, eventgrid). Adding 3.13 to unit tests is pending still -- we don't want to run proxy worker tests for <=3.12, and vice versa. This will involve some more logic

tests/extension_tests/deferred_bindings_tests/test_deferred_bindings.py
```diff
@@ -77,9 +77,9 @@ async def test_deferred_bindings_enabled_log(self):
 
 @unittest.skipIf(sys.version_info.minor <= 8, "The base extension"
                                               "is only supported for 3.9+.")
-@unittest.skipIf(sys.version_info.minor < 13, "For python 3.13+,"
-                                              "this logic is in the"
-                                              "library worker.")
+@unittest.skipIf(sys.version_info.minor >= 13, "For python 3.13+,"
```
swapnil-nagar, line 80: Use a constant to defined the supported python version to minimize the changes in the future.

eng/templates/jobs/ci-unit-tests.yml
```diff
@@ -35,19 +35,22 @@ jobs:
...
-          echo "Python version from variable: $PYTHON_VERSION"
-          if (( $(echo "$PYTHON_VERSION < 3.13" | bc -l) )); then
-              echo "Running unittests (Python < 3.13)..."
+          PY_VER="$(PYTHON_VERSION)"
+          echo "Python version: $PY_VER"
+      
+          # Extract minor version as integers
+          PY_MINOR="${PY_VER#*.}"
+      
+          if [ "$PY_MINOR" -ge 13 ]  then
```
swapnil-nagar: NIT: Consider having a file that contains the supported version of python and used at both ci-unit-tests and ci-e2e-tests.yml file.

tests/extension_tests/deferred_bindings_tests/test_deferred_bindings.py
```diff
@@ -43,9 +43,9 @@ def __init__(self, version: str, source: str,
...
```
swapnil-nagar, line 46: [Checking] You are skipping tests for 3.13 is that intended?

eng/templates/jobs/ci-unit-tests.yml
```diff
@@ -35,19 +35,22 @@ jobs:
...
+              echo "Running proxy_worker tests (Python >= 3.13)..."
```
swapnil-nagar, line 45: NIT: Add the message with specific Python version for which you are running the tests.

echo "Running proxy_worker tests for Python $(PYTHON_VERSION).$(PY_MINOR)"

gavin-aguiar: This pipeline will be run for all versions greater than 3.13. Adding the major version in the message doesn't help much since the major version won't change

eng/templates/official/jobs/ci-e2e-tests.yml
vrdmr: >E2E tests are re-using 312 resources for now until the appropriate 313 resources are created (cosmosdb, sql, eventgrid).

For this - is there a helper script to smoothen the process of creation? If not, please add it in the 3.13 backlog.

eng/templates/official/jobs/build-artifacts.yml
```diff
@@ -100,6 +106,9 @@ jobs:
       Python312V4:
```
vrdmr, line 106: In the long run - We should simplify these blocks - the yaml is unnecessarily long.

eng/templates/official/jobs/ci-e2e-tests.yml
```diff
@@ -129,7 +145,6 @@ jobs:
             Write-Host "##vso[task.setvariable variable=skipTest;]false"
           }
         displayName: 'Set skipTest variable'
-        condition: or(eq(variables.isSdkRelease, true), eq(variables['USETESTPYTHONSDK'], true))
```
vrdmr, line 132: What did this do?

pack/scripts/mac_arm64_deps.sh
```diff
@@ -13,4 +13,10 @@ python -m invoke -c test_setup build-protos
...
 cd ..
 cp .artifactignore "$BUILD_SOURCESDIRECTORY/deps"
-cp -r azure_functions_worker/protos "$BUILD_SOURCESDIRECTORY/deps/azure_functions_worker"
\ No newline at end of file
+
+version_minor=$(echo $1 | cut -d '.' -f 2)
+if [[ $version_minor -lt 13 ]]; then
+    cp -r azure_functions_worker/protos "$BUILD_SOURCESDIRECTORY/deps/azure_functions_worker"
+else
+    cp -r proxy_worker/protos "$BUILD_SOURCESDIRECTORY/deps/proxy_worker"
```
vrdmr, line 21: in the long run - the name shouldn't change - can it be done in this iteration itself?

vrdmr: As this seems unnecessary.

tests/utils/testutils.py
vrdmr: I really don't like this :( Better to create a 3.13 specific util for now.

tests/unittests/test_code_quality.py
```diff
@@ -43,7 +43,8 @@ def test_flake8(self):
 
         try:
             subprocess.run(
-                [sys.executable, '-m', 'flake8', '--config', str(config_path)],
+                [sys.executable, '-m', 'flake8', '--config', str(config_path),
+                 'azure_functions_worker',],
```
vrdmr, line 47: Would it not "flake" other folders?

python/test/worker.py
vrdmr: We wanted to get out of the whole minor version thing - Please create a different test folder if needed.

python/proxyV4/worker.py
vrdmr, line 1: Why create a separate worker.py?

python/proxyV4/worker.py
vrdmr: This needs a documentation update

python/prodV4/worker.config.json
```diff
@@ -3,7 +3,7 @@
         "language":"python",
         "defaultRuntimeVersion":"3.11",
```
vrdmr: change default as well - to 12

proxy_worker/version.py
vrdmr: Any specific reason it should be in a python file or would it be in the pyproject.toml itself

tests/extension_tests/deferred_bindings_tests/test_deferred_bindings.py
```diff
@@ -6,8 +6,9 @@
 import azure.functions as func
 from tests.utils import testutils
 
-from azure_functions_worker import protos
-from azure_functions_worker.bindings import datumdef, meta
+if sys.version_info.minor < 13:
```
vrdmr, line 9: Still sad :(

proxy_worker/start_worker.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Python Azure Functions Worker')
+    parser.add_argument('--host',
+                        help="host address")
+    parser.add_argument('--port', type=int,
+                        help='port number')
+    parser.add_argument('--workerId', dest='worker_id',
```
vrdmr, line 17: Clean up unused args

vrdmr: https://github.com/Azure/azure-functions-host/blob/7e3e7b78fa90ec18eaefd995a9e7797d326d2d00/src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs#L45

proxy_worker/start_worker.py
```diff
@@ -0,0 +1,83 @@
...
+    from . import logging
+    from .logging import error_logger, logger
+
+    args = parse_args()
+    logging.setup(log_level=args.log_level, log_destination=args.log_to)
+
+    logger.info("Args: %s", args)
+    logger.info('Starting proxy worker.')
+    logger.info('Worker ID: %s, Request ID: %s, Host Address: %s:%s',
+                args.worker_id, args.request_id, args.host, args.port)
```
vrdmr: Change to 1 or debug logs

proxy_worker/start_worker.py
```diff
@@ -0,0 +1,83 @@
...
+        raise
+
+
+async def start_async(host, port, worker_id, request_id):
+    from . import dispatcher
+
+    disp = await dispatcher.Dispatcher.connect(host=host, port=port,
+                                               worker_id=worker_id,
+                                               request_id=request_id,
+                                               connect_timeout=5.0)
```
vrdmr: not urgent - Change the 5.0 to a CONSTANT or a value thats passed.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+import asyncio
+import concurrent.futures
+import logging
+import os
+import queue
+import sys
+import threading
+import traceback
+import typing
+from asyncio import AbstractEventLoop
```
vrdmr, line 13: isort this whole project.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+            task_invocation_id = getattr(current_task,
+                                         ContextEnabledTask.AZURE_INVOCATION_ID,
+                                         None)
+            if task_invocation_id is not None:
+                return task_invocation_id
+
+    return getattr(_invocation_id_local, 'invocation_id', None)
+
+
+class AsyncLoggingHandler(logging.Handler):
```
vrdmr, line 78: Exploration for later - to move to logging

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+                request_handler = self._handle__worker_init_request
+            case "function_environment_reload_request":
+                request_handler = self._handle__function_environment_reload_request
+            case "functions_metadata_request":
+                request_handler = self._handle__functions_metadata_request
+            case "function_load_request":
+                request_handler = self._handle__function_load_request
+            case "worker_status_request":
+                request_handler = self._handle__worker_status_request
+            case "invocation_request":
```
vrdmr, line 257: make invocation request first - as it's the most used eventually.

vrdmr: Or take a look at match case in Python - and see if ordering has an impact.

proxy_worker/dispatcher.py
vrdmr: > the new logs added are intended to help confirm that the library worker is receiving requests appropriately.

No new info logs - make them debug. During cold start, adding a log should be fine but post Evan's validation.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+                    sys.version,
+                    VERSION,
+                    self.request_id)
+
+        if DependencyManager.should_load_cx_dependencies():
+            DependencyManager.prioritize_customer_dependencies()
+
+        global _library_worker
+        directory = request.worker_init_request.function_app_directory
+        v2_directory = os.path.join(directory, get_script_file_name())
```
vrdmr: Rename - make it explicit that its a function_app entry file for v2 model

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+        self._grpc_connected_fut = loop.create_future()
+        self._grpc_thread: Optional[threading.Thread] = threading.Thread(
+            name='grpc_local-thread', target=self.__poll_grpc)
+
+        self._sync_call_tp: Optional[concurrent.futures.Executor] = (
+            self._create_sync_call_tp(self._get_sync_tp_max_workers()))
+
+    def on_logging(self, record: logging.LogRecord,
+                   formatted_msg: str) -> None:
+        if record.levelno >= logging.CRITICAL:
```
swapnil-nagar, line 143: Thats a lot uses an if-elif chain to map log levels, consider having a map defined to clean the code.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+                return
+            error_logger.exception(
+                'unhandled error in gRPC thread. Exception: {0}'.format(
+                    ''.join(traceback.format_exception(ex))))
+            raise
+
+    async def _dispatch_grpc_request(self, request):
+        content_type = request.WhichOneof("content")
+
+        match content_type:
```
swapnil-nagar, line 246: Instead of if-else, use a dictionary

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+
+            self._loop.set_task_factory(self._old_task_factory)
+            self.stop()
+
+    def stop(self) -> None:
+        if self._grpc_thread is not None:
+            self._grpc_resp_queue.put_nowait(self._GRPC_STOP_RESPONSE)
+            self._grpc_thread.join()
+            self._grpc_thread = None
+
```
swapnil-nagar, line 328: [Blocker] You are joining thread and cleanup, there could be exception from these which will left resources in open(or undefined state).  I will encourage use to exception and finally to ensure thread safety.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+        func_env_reload_request = \
+            request.function_environment_reload_request
+        directory = func_env_reload_request.function_app_directory
+        DependencyManager.prioritize_customer_dependencies(directory)
+
+        global _library_worker
+        v2_directory = os.path.join(directory, get_script_file_name())
+        if os.path.exists(v2_directory):
+            try:
+                import azure_functions_worker_v2  # NoQA
```
vrdmr: Explore to import this in a separate variable above.

x = importlib.import (azure_functions_worker_v2)

x.handle_dispatch_init in your code.

Decide if you want to run import lib again based on a return `DependencyManager.prioritize_customer_dependencies(directory)` if it tells you that Cx did bring in a runtime in reqs.txt

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+            raise RuntimeError('there can be only one running dispatcher per '
+                               'process')
+
+        self._old_task_factory = self._loop.get_task_factory()
+
+        DispatcherMeta.__current_dispatcher__ = self
+        try:
+            forever = self._loop.create_future()
+
+            self._grpc_resp_queue.put_nowait(
```
swapnil-nagar, line 278: [Checking, unaware of python data structures] If multiple thread attempt to put in the queue is that thread safe. We use something called BlockingQueue or ConcurrentQueue in Java

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+                logger.warning('Switched to console logging due to exception.')
+        finally:
+            DispatcherMeta.__current_dispatcher__ = None
+
+            self._loop.set_task_factory(self._old_task_factory)
+            self.stop()
+
+    def stop(self) -> None:
+        if self._grpc_thread is not None:
+            self._grpc_resp_queue.put_nowait(self._GRPC_STOP_RESPONSE)
```
swapnil-nagar, line 325: Also you should consider having lock, if two thread access this code simultaneously, it will lead to race cndition.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+                if msg is self._GRPC_STOP_RESPONSE:
+                    grpc_req_stream.cancel()
+                    return
+                yield msg
+
+        grpc_req_stream = stub.EventStream(gen(self._grpc_resp_queue))
+        try:
+            for req in grpc_req_stream:
+                self._loop.call_soon_threadsafe(
+                    self._loop.create_task, self._dispatch_grpc_request(req))
```
swapnil-nagar, line 233: [Checking] If _dispatch_grpc_request itself is thread safe? You might need to synchronize the _dispatch_grpc_request

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+            invocation_id = getattr(
+                current_task, self.AZURE_INVOCATION_ID, None)
+            if invocation_id is not None:
+                self.set_azure_invocation_id(invocation_id)
+
+    def set_azure_invocation_id(self, invocation_id: str) -> None:
+        setattr(self, self.AZURE_INVOCATION_ID, invocation_id)
+
+
+_invocation_id_local = threading.local()
```
swapnil-nagar, line 61: [Checking] Why local? if application scale to different threadpool, will local storage be an issue and lead to thread safety?

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,528 @@
...
+        self._loop = loop
+        self._host = host
+        self._port = port
+        self._request_id = request_id
+        self._worker_id = worker_id
+        self._grpc_connect_timeout: float = grpc_connect_timeout
+        self._grpc_max_msg_len: int = grpc_max_msg_len
+        self._old_task_factory: Optional[Any] = None
+
+        self._grpc_resp_queue: queue.Queue = queue.Queue()
```
swapnil-nagar, line 133: Consider having a blocking queue equivalent in python, as it is accessed by multiple threads.

proxy_worker/logging.py
```diff
@@ -0,0 +1,92 @@
...
+    error_logger.setLevel(getattr(logging, log_level))
+
+
+def disable_console_logging() -> None:
+    # We should only remove the sys.stdout stream, as error_logger is used for
+    # unexpected critical error logs handling.
+    if logger and handler:
+        handler.flush()
+        logger.removeHandler(handler)
+
```
swapnil-nagar, line 73: [Checking] Should you need close  handler?

proxy_worker/dispatcher.py
gavin-aguiar: Python is single threaded. Shouldn't have this issue

proxy_worker/dispatcher.py
gavin-aguiar: This has not related to storage. Its a  variable used to store the data of the current thread

proxy_worker/dispatcher.py
gavin-aguiar: In match-case the order of the case statements does not impact performance because the pattern matching is compiled into efficient jump tables or lookups internally by the interpreter.

proxy_worker/dispatcher.py
vrdmr: Can this be switch case itself?

proxy_worker/start_worker.py
```diff
@@ -0,0 +1,77 @@
...
+    from . import logging
+    from .logging import error_logger, logger
+
+    args = parse_args()
+    logging.setup(log_level="INFO", log_destination=None)
+
+    logger.info("Args: %s", args)
+    logger.info(
+        'Starting proxy worker. Worker ID: %s, Request ID: %s, Host Address: %s:%s',
+        args.worker_id, args.request_id, args.host, args.port)
```
vrdmr, line 54: Merging these logs

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,311 @@
...
+            cls.worker_deps_path, cx_deps_path, working_directory,
+            DependencyManager.is_in_linux_consumption(),
+            is_envvar_true("WEBSITE_PLACEHOLDER_MODE"), sys.path)
+
+        cls._remove_from_sys_path(cls.worker_deps_path)
+        cls._add_to_sys_path(cls.worker_deps_path, True)
+        cls._add_to_sys_path(cls.cx_deps_path, True)
+        cls._add_to_sys_path(working_directory, False)
+
+        logger.info(f'Finished prioritize_customer_dependencies: {sys.path}')
```
vrdmr, line 151: Please convert to debug (when GA)

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,311 @@
...
+            'working_directory: %s, Linux Consumption: %s, Placeholder: %s, '
+            'sys.path: %s',
+            cls.worker_deps_path, cx_deps_path, working_directory,
+            DependencyManager.is_in_linux_consumption(),
+            is_envvar_true("WEBSITE_PLACEHOLDER_MODE"), sys.path)
+
+        cls._remove_from_sys_path(cls.worker_deps_path)
+        cls._add_to_sys_path(cls.worker_deps_path, True)
+        cls._add_to_sys_path(cls.cx_deps_path, True)
+        cls._add_to_sys_path(working_directory, False)
```
vrdmr, line 149: Can adding to first or last be more explicit - readable

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,311 @@
...
+
+        logger.info(f'Finished prioritize_customer_dependencies: {sys.path}')
+
+    @classmethod
+    def _add_to_sys_path(cls, path: str, add_to_first: bool):
```
vrdmr, line 154: move private methods to be earlier than public methods.

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,311 @@
...
+            Linux Dedicated/Premium: path to customer's site packages
+            Linux Consumption: empty string
+        """
+        prefix: Optional[str] = os.getenv(AZURE_WEBJOBS_SCRIPT_ROOT)
+        cx_paths: List[str] = [
+            p for p in sys.path
+            if prefix and p.startswith(prefix) and ('site-packages' in p)
+        ]
+        # Return first or default of customer path
+        return (cx_paths or [''])[0]
```
vrdmr, line 231: Why is "AZURE_WEBJOBS_SCRIPT_ROOT" showing up here?

vrdmr: First - explore using the dir from grpc request itself. or use cls.Cx_working_directory

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,311 @@
...
+        """Get the worker dependency sys.path. This will always available
+        even in all skus.
+
+        Returns
+        -------
+        str
+            The worker packages path
+        """
+        # 1. Try to parse the absolute path python/3.13/LINUX/X64 in sys.path
+        r = re.compile(r'.*python(\/|\\)\d+\.\d+(\/|\\)(WINDOWS|LINUX|OSX).*')
```
vrdmr, line 257: This Regex should be on the top as Constant

proxy_worker/utils/dependency.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        if not path:
+            return
+
+        not_builtin = set(sys.modules.keys()) - set(sys.builtin_module_names)
+
+        # Don't reload proxy_worker
+        to_be_cleared_from_cache = set([
+            module_name for module_name in not_builtin
+            if not module_name.startswith('proxy_worker')
```
vrdmr, line 286: Name - shouldn't be hardcoded. Should be moved to Constant.

proxy_worker/utils/constants.py
```diff
@@ -0,0 +1,15 @@
...
+PYTHON_ENABLE_DEBUG_LOGGING = "PYTHON_ENABLE_DEBUG_LOGGING"
+PYTHON_THREADPOOL_THREAD_COUNT = "PYTHON_THREADPOOL_THREAD_COUNT"
```
vrdmr, line 6: Proxy worker shouldn't create the TP. Thus this constant shouldn't be here.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,521 @@
...
+        self._grpc_connected_fut = loop.create_future()
+        self._grpc_thread: Optional[threading.Thread] = threading.Thread(
+            name='grpc_local-thread', target=self.__poll_grpc)
+
+        self._sync_call_tp: Optional[concurrent.futures.Executor] = (
+            self._create_sync_call_tp(self._get_sync_tp_max_workers()))
+
```
vrdmr, line 140: We should explore - moving this logic into lib_worker and not to put in proxy worker.

proxy_worker/dispatcher.py
```diff
@@ -0,0 +1,521 @@
...
+        return self._request_id
+
+    @property
+    def worker_id(self) -> str:
+        return self._worker_id
+
+    @classmethod
+    async def connect(cls, host: str, port: int, worker_id: str,
+                      request_id: str, connect_timeout: float):
+        loop = asyncio.events.get_event_loop()
```
vrdmr, line 188: Add a task - app setting for enabling "uvloop" - add this feature.

`uvloop.new_event_loop` vs `asyncio.events.get_event_loop()`

proxy_worker/dispatcher.py
vrdmr: So something like

```python
handlers = handlers["content_type"]
```

And handlers is a dict [<content_type:str>, <func_point:addr>]

proxy_worker/dispatcher.py
vrdmr: >which will left resources in open(or undefined state)

This is already at the end of the workers lifetime - and we can discard any state.

vrdmr (approved):
Looks good to go in;

follow up tasks to be created and minor tests.

PR #1700, opened 2025-06-13
fix: validate_function_params error message

Kisbogyi:
## Description

Fixes #
function argument error message
---
When somebody created a trigger with wrong parameters the error said: "FunctionLoadError: cannot load the healthcheck function: the following parameters are declared in function.json but not in Python: {'myTimer'}" but with python function v2 functin.json is not utilized.

Examle code snippet to reproduce issue: 

```py
@bp_healthcheck.timer_trigger(
    schedule="0 */30 * * * *", arg_name="myTimer"
)
async def healthcheck() -> None:
```

Solution is that I rewrote the wording of the exception.

There is this comment in the call stack so I think `validate_function_params` should not be called when v1 function is loaded. (azure_functions_worker/dispatcher.py line 457)

```
# For V1, the function path will not exist and
# return None.
```
Moreover it has an annotations parameter, which is v2 only. So this should not be a problem.

Tests were updated!

### PR information

### DIFF
--- a/azure_functions_worker/functions.py
+++ b/azure_functions_worker/functions.py
@@ -136,14 +136,14 @@ class Registry:
         if set(params) - set(bound_params):
             raise FunctionLoadError(
                 func_name,
-                'the following parameters are declared in Python but '
-                f'not in function.json: {set(params) - set(bound_params)!r}')
+                'the following parameters are declared in function parameter but '
+                f'not in trigger annotation: {set(params) - set(bound_params)}')
 
         if set(bound_params) - set(params):
             raise FunctionLoadError(
                 func_name,
-                f'the following parameters are declared in function.json but '
-                f'not in Python: {set(bound_params) - set(params)!r}')
+                f'the following parameters are declared in trigger annotation but '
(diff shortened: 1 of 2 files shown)


hallvictoria:
Hi @Kisbogyi, thanks for this. I agree this error message is misleading when using the V2 programming model. This is definitely something we can get fixed.

`validate_function_params` is used by both the V1 & V2 programming model though, so we'll want to ensure that the error message is understandable regardless of programming model. 

(For V1, the function is validated in `_handle__function_load_request` through the `add_function` call. That comment refers to indexing in `_handle__functions_metadata_request`, which is for V2. Also, the "annotations" parameter refers to when arguments are defined with type hints, like `req: func.HttpRequest`)

Kisbogyi:
Hi!
Sorry for the misunderstanding, I should have understand the code better before posting the pr. 
Should I open an issue for this?
Sorry, for your time, and thanks!

hallvictoria:
No worries! This is a good point to raise. Yes, please open an issue, and we'll start looking into it

PR #1763, opened 2025-09-08
feat: support sb settlement client

hallvictoria:
## Description

**Proposed design (cx view)**
image

Customers will add a second parameter with the type `ServiceBusMessageActions` - no associated binding. That param will return a grpc client type that will handle the settlement commands.

**Flow of worker changes:**

1. Indexing: identify if there is a param with no associated binding
   - If yes: check if it's a settlement client via check_grpc_client_type - base extension method
      - If yes: pop arg from params - we will not do any more validation on it (nothing needed)
      - If no: app is broken, we fail
   - If no: pass
2. Invocation: return the grpc client
   - We saved the corresponding arg name for the client in a param
   - We get the client via methods exposed in the base and return it to the arg

Fixes #

---

### Host-Worker Contract
   - If no, please explain why:   

### Worker Execution Logic
- [x] Does this PR affect worker execution logic (e.g., function invocation, bindings, lifecycle)?

**Python Version Coverage**
   - [x] Does this change apply to both Python <=3.12 and 3.13+?
      - [x] azure_functions_worker (Python <= 3.12)
      - [x] runtimes/v1 / runtimes/v2 (Python >= 3.13)
   - If no, please explain why:

   - [x] V2 programming model (runtimes/v2)?


workers/azure_functions_worker/bindings/meta.py
```diff
@@ -314,3 +313,33 @@ def get_deferred_raw_bindings(indexed_function, input_types):
     raw_bindings, bindings_logs = DEFERRED_BINDING_REGISTRY.get_raw_bindings(
         indexed_function, input_types)
     return raw_bindings, bindings_logs
+
+
+def get_settlement_client():
+    return DEFERRED_BINDING_REGISTRY.get("serviceBusClient").get_client()
```
hallvictoria: todo: make this a constant in the base?

workers/azure_functions_worker/bindings/meta.py
```diff
@@ -314,3 +313,33 @@ def get_deferred_raw_bindings(indexed_function, input_types):
...
+    """
+    for missing_param in set(params) - set(bound_params):
+        try:
+            param_type = annotations.get(missing_param)
+            settlement_client_enabled = DEFERRED_BINDING_REGISTRY.check_grpc_client_type(
+                param_type)
+            if settlement_client_enabled:
+                return (settlement_client_enabled, missing_param)
+        except Exception:
+            param_type = None
```
hallvictoria: todo: break the loop here. if an app has an arg without an associated binding that's not a settlement client, it's a broken app

runtimes/v2/azure_functions_runtime/bindings/meta.py
```diff
@@ -286,3 +287,39 @@ def get_deferred_raw_bindings(indexed_function, input_types):
...
+    if len(set(params) - set(bound_params)) > 1:
+        return ''
+
+    # There is only one unbound param, check the type
+    settlement_param = next(iter(set(params) - set(bound_params)))
+    try:
+        param_type = annotations.get(settlement_param)
+        # Check if the type is a supported type for the settlement client
+        if DEFERRED_BINDING_REGISTRY.check_supported_grpc_client_type(param_type):
+            return settlement_param
```
hallvictoria, line 322: Raising awareness for this case - if the cx has a bad app & pins to an older version of the base, this check will fail and say the base has no method called `check_supported_grpc_client`. IMO this is okay - cx should not pin base and it is already a bad app

runtimes/v2/azure_functions_runtime/bindings/meta.py
```diff
@@ -286,3 +287,39 @@ def get_deferred_raw_bindings(indexed_function, input_types):
...
+    except Exception:
+        param_type = None
+    return ''
```
gavin-aguiar: Can we return None instead?

hallvictoria: Done

gavin-aguiar (approved):

PR #1857, opened 2026-04-24
feat: proxy worker changes to support the runtime base

hallvictoria:
## Description

1. Each Runtime Declares an Entry Point
```
[project.entry-points."azurefunctions.runtimes"]
fastapi = "azure_functions_fastapi:Runtime"
```
2. Dispatcher auto-discovers all installed runtimes

- Zero worker changes for new runtimes
- No hardcoded list
- Lazy loading - only imports what's installed
- Light-weight metadata scan: reads .dist-info/METADATA or .egg-info/PKG-INFO files, typically milliseconds even with lots of packages

### Alternative approach - identify & import any packages with matching name pattern
```
def discover_installed_runtimes():
    """
    Discover all installed azure_functions_* runtime packages.
    No hardcoded list needed - uses package name convention.
    """
    import pkgutil
    
    runtime_packages = []
    for importer, modname, ispkg in pkgutil.iter_modules():
        # Match naming convention: azure_functions_<framework>
        if (modname.startswith('azure_functions_') and 
            modname not in ['azure_functions', 'azure_functions_worker']):
            runtime_packages.append(modname)
    
    return runtime_packages

################################################################
for runtime_package in discover_installed_runtimes():
    try:
        importlib.import_module(runtime_package)  # Triggers metaclass registration
    except ImportError:
        continue

if RuntimeFeatureChecker.runtime_loaded():
    # Use the registered runtime
```

Fixes #

---

### Host-Worker Contract
   - If no, please explain why:   

### Worker Execution Logic

**Python Version Coverage**
   - If no, please explain why:


workers/proxy_worker/dispatcher.py
```diff
@@ -417,7 +420,76 @@ def stop(self) -> None:
...
+                # Load the single runtime entry point if available
+                if available_runtimes:
+                    ep = available_runtimes[0]
+                    try:
+                        # Load the entry point (triggers import and
+                        # metaclass registration)
+                        ep.load()
+                        logger.debug(f"Loaded runtime entry point: {ep.name}")
+                    except Exception as e:
+                        logger.debug(f"Could not load runtime {ep.name}: {e}")
```
gavin-aguiar: Should we throw and return after this?

workers/proxy_worker/dispatcher.py
```diff
@@ -417,7 +420,76 @@ def stop(self) -> None:
...
+                                     runtime_name, runtime_module_name, package_name)
+
+                        # Import the top-level runtime package (which exports
+                        # the public API)
+                        runtime_module = importlib.import_module(package_name)
+                        _library_worker = runtime_module
+                        _library_worker_has_cv = _library_worker.invocation_id_cv
+                        # Module has been imported, end check
+                        return
+            except ImportError:
```
gavin-aguiar: Do we need a broader exception handling too?

workers/proxy_worker/dispatcher.py
hallvictoria: Done!

workers/proxy_worker/dispatcher.py
hallvictoria: Added try-catch around the whole block. If at any point we have an exception occur (import, loading, multiple runtimes, etc), we catch & then raise the error

workers/proxy_worker/dispatcher.py
```diff
@@ -417,7 +420,87 @@ def stop(self) -> None:
...
+                    )
+
+                # Load the single runtime entry point if available
+                if available_runtimes:
+                    ep = available_runtimes[0]
+                    try:
+                        # Load the entry point (triggers import and
+                        # metaclass registration)
+                        ep.load()
+                        logger.debug(f"Loaded runtime entry point: {ep.name}")
```
gavin-aguiar: Nit: Don't use fstrings for logging.

workers/proxy_worker/dispatcher.py
```diff
@@ -417,7 +420,87 @@ def stop(self) -> None:
...
+                    except Exception as e:
+                        raise RuntimeError(
+                            f"Failed to load runtime entry point {ep.name}: {e}"
+                        )
+
+                    # Check if a runtime was registered
+                    # Check if the runtime base package has the RuntimeFeatureChecker
+                    # Check if the runtime is loaded
+                    if runtime_base is not None \
+                        and hasattr(runtime_base, 'RuntimeFeatureChecker') \
```
gavin-aguiar, line 472: If the worker has an older version of the base, it won't go inside this if condition, and it will fail silently. We can either throw or add an error log in the else condition.

workers/proxy_worker/dispatcher.py
```diff
@@ -417,7 +420,87 @@ def stop(self) -> None:
...
+                        # Import the top-level runtime package (which exports
+                        # the public API)
+                        runtime_module = importlib.import_module(package_name)
+                        _library_worker = runtime_module
+                        _library_worker_has_cv = _library_worker.invocation_id_cv
+
+                        # Module has been imported, end check
+                        return
+            except Exception as e:
+                logger.info("Error when loading runtime: %s",
```
gavin-aguiar: logger.error

workers/proxy_worker/dispatcher.py
```diff
@@ -417,7 +420,87 @@ def stop(self) -> None:
...
+                            traceback.format_exc())
+                raise e
+
+        # No runtime registered via base package
+        # Use traditional detection
+        logger.debug(
```
gavin-aguiar: This log will be noisy even though is a debug log. Do we need this since we already have debug logs in the if condition?

workers/proxy_worker/dispatcher.py
hallvictoria: Added an else statement to log & throw an error for unsupported base version

gavin-aguiar (approved):

PR #1874, opened 2026-06-09
test: Test fixes

gavin-aguiar:
## Description
Fixes the ServiceBus SDK binding failing on the classic worker (Python ≤ 3.12), which caused emulator tests to return  404  and function apps using the extension to fail indexing. The root cause was the worker's protobuf vendoring shadowing the app's newer protobuf.

### Version-aware protobuf vendoring:

• The worker now inspects whichever  google.protobuf  is first on  sys.path  (via  find_spec ) and activates the vendored copy only when the app's protobuf is older than the vendored version. Same-or-newer protobuf (including 6.x) is used directly.
• Result: protobuf-6 extensions load, the worker keeps fast  upb , and older customer protobuf is still shadowed for safety.

####  Launchers:

• Removed the forced  _AZFUNC_USE_VENDORED_PROTOBUF=1  override so the worker autodetects. Local Core Tools now matches production behavior.

#### Uniform top-level imports:

• Worker pb2 stubs,  loader.py , and  nullable_converters.py  all import top-level  google.protobuf  so the entire worker runs in one descriptor pool.  REWRITE_PROTOBUF=False  keeps generated stubs top-level.

Pack script updates ( eng/pack/templates/{win,nix,nix_arm64,macos_64}_env_gen.yml )

Because the worker now imports the top-level  google.protobuf , the packaged artifact must include it:

• Removed the  !google/**  exclusion from the classic-worker copy step (it was dropping the top-level  google  package, causing  ModuleNotFoundError: No module named 'google'  at startup when an app shipped no protobuf).
• Added  google  back to the classic-worker validation list so packaging fails fast if it's ever missing.

### Performance

• Switch pip → uv in all install scripts and CI templates ( install-dependencies.sh ,  test-sdk.sh ,  test-extensions.sh ,  ci-dependency-check.yml ,  ci-fc-tests.yml ) for faster dependency resolution. Batched the worker installs into a single resolver pass.
• Enable real pytest-xdist parallelism: unit tests  -n auto , emulator tests  -n 2 , E2E kept serial (shared Azure resources).  --dist loadfile  alone was previously a no-op without  -n .
• Build only the WebHost project instead of the whole  WebJobs.Script.sln  (~2.7x faster, far less disk, avoids restoring unrelated projects from the internal feed). Added  set -e  to  test-setup.sh  so build failures surface instead of being masked.
• Removed unused dev deps ( ptvsd ,  pre-commit ,  pytest-sugar ,  pytest-randomly ) from all three  pyproject.toml .
• Split the install step into separate "Python deps (uv)" and "webhost build (dotnet)" stages so timings are visible.

### Test reliability

• Fixed flaky  test_unhandled_error  (all 3 variants): the traceback is forwarded to the host asynchronously, so the log-check snapshot raced it. Added  WebHostTestCase.wait_for_host_log()  to wait for the line before asserting.
• Replaced the racy  time.sleep(10)  webhost startup with a real readiness probe polling  /admin/host/status  for  Running  (with master key).
• Fixed Azurite "InvalidHeaderValue" in emulator tests via  --skipApiVersionCheck --loose ; added a ServiceBus emulator readiness wait.

### Python 3.14 / version compatibility

• Made  runtimes/v{1,2}  installs and the uamqp-based ServiceBus binding install conditional (uv enforces  requires-python ); loosened the  fastapi  pin to resolve a  dev  vs  test-http-v2  conflict.
• Gated  TestServiceBusSDKFunctions  to Python 3.13 only (the binding needs protobuf ≥6.32/grpcio ≥1.74, incompatible with the ≤3.12 worker's pins).

### New coverage

• Added a Flex Consumption test verifying blob SDK (deferred) bindings load under Flex ( BlobSdkBindings.zip  fixture +  test_blob_sdk_bindings , gated to 3.14).

Fixes #

---

### Host-Worker Contract
   - If no, please explain why:   

### Worker Execution Logic

**Python Version Coverage**
   - If no, please explain why:


eng/templates/jobs/ci-library-unit-tests.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           eng/scripts/install-dependencies.sh $(PYTHON_VERSION) ${{ parameters.PROJECT_DIRECTORY }}
         displayName: 'Install dependencies'
       - bash: |
-          python -m pytest -q --dist loadfile --reruns 4 --instafail --cov=./${{ parameters.PROJECT_DIRECTORY }} --cov-report xml --cov-branch tests/unittests
+          python -m pytest -q -n auto --dist loadfile --reruns 4 --reruns-delay 5 --instafail --cov=./${{ parameters.PROJECT_DIRECTORY }} --cov-report xml --cov-branch tests/unittests
```
hallvictoria, line 52: Should `n` be auto or 2? In ci-emulator-tests, it's 2

eng/templates/jobs/ci-emulator-tests.yml
```diff
@@ -128,7 +130,7 @@ jobs:
           docker compose -f ${{ parameters.PROJECT_DIRECTORY }}/tests/emulator_tests/utils/eventhub/docker-compose.yml up -d
         displayName: 'Install Azurite and Start EventHub Emulator'
       - bash: |
-            python -m pytest -q --dist loadfile --reruns 4 --ignore=tests/emulator_tests/test_servicebus_functions.py tests/emulator_tests
+            python -m pytest -q -n 2 --dist loadfile --reruns 4 --ignore=tests/emulator_tests/test_servicebus_functions.py tests/emulator_tests
```
hallvictoria, line 133: `--reruns-delay 5`?

runtimes/v2/tests/unittests/test_deferred_bindings.py
```diff
@@ -24,6 +26,7 @@ def setUp(self):
         # Initialize DEFERRED_BINDING_REGISTRY
         meta.load_binding_registry()
 
+    @unittest.skip("TODO: Move to emulator.")
```
hallvictoria: We can move this test to the other SDK bindings tests then

workers/tests/emulator_tests/test_servicebus_functions.py
```diff
@@ -67,7 +67,12 @@ def get_script_dir(cls):
             'servicebus_functions_stein' / 'generic'
 
 
-@unittest.skipIf(sys.version_info.minor >= 14, "Skip to figure out uamqp.")
+@unittest.skipUnless(
```
hallvictoria: Can we keep the existing structure for installing protobuf / grpcio for these tests? Ideally, we should test <= 3.12 to make sure we're also testing azure_functions_worker

workers/tests/unittests/test_third_party_http_functions.py
```diff
@@ -134,9 +134,13 @@ def test_return_http_no_body(self):
             self.assertEqual(r.text, '')
             self.assertEqual(r.status_code, 200)
 
+        @testutils.retryable_test(3, 5)
         def test_unhandled_error(self):
             r = self.webhost.request('GET', 'unhandled_error', no_prefix=True)
             self.assertEqual(r.status_code, 500)
+            # The worker forwards the traceback to the host asynchronously, so
+            # wait for it before check_log_unhandled_error reads its snapshot.
+            self.wait_for_host_log('ZeroDivisionError: division by zero')
```
hallvictoria, line 143: Are these 3 `test_unhandled_error` tests duplicates of each other?

workers/tests/utils/testutils.py
```diff
@@ -117,6 +117,10 @@
 }
 """
 
+# Master key defined in SECRETS_TEMPLATE above; required to call the
+# host's protected /admin endpoints (e.g. /admin/host/status).
+MASTER_KEY = "testMasterKey"
```
hallvictoria: Can you move this to `workers\tests\utils\constants.py`?

workers/pyproject.toml
```diff
@@ -54,29 +54,25 @@ dev = [
...
     "pytest-rerunfailures",
     "pytest-asyncio",
-    "ptvsd",
     "python-dotenv",
     "plotly",
     "scikit-learn",
     "opencv-python",
     "pandas",
     "numpy",
-    "pre-commit",
```
hallvictoria, line 79: Was `pre-commit` used anywhere? I vaguely remember a PR adding it

workers/pyproject.toml
```diff
@@ -54,29 +54,25 @@ dev = [
...
     "coverage",
-    "pytest-sugar",
     "opentelemetry-api", # Used for OpenTelemetry unit tests
     "pytest-cov",
     "pytest-xdist",
-    "pytest-randomly",
     "pytest-instafail",
     "pytest-rerunfailures",
     "pytest-asyncio",
-    "ptvsd",
```
hallvictoria, line 72: Do we need this for VSCode debugging? If someone is debugging the worker locally? I forget how exactly this package is used

eng/templates/jobs/ci-library-unit-tests.yml
gavin-aguiar: The auto is correct here since these tests run independently  and do not compete for resources. Emulator tests  uses  -n 2  specifically because they share stateful external resources

eng/templates/jobs/ci-emulator-tests.yml
gavin-aguiar: We dont need --reruns-delay  here. The emulator tests already handle transient timing with wait_until_ready and time.sleep

workers/tests/unittests/test_third_party_http_functions.py
gavin-aguiar: Nope. They check v1, streaming and wsgi

workers/pyproject.toml
gavin-aguiar: This was used for the . pre-commit-config.yaml but even this is not used so removed the yaml file too.

workers/pyproject.toml
gavin-aguiar: Nope this is not used. For vscode we use debugpy

runtimes/v2/tests/unittests/test_deferred_bindings.py
gavin-aguiar: moved.

workers/tests/utils/testutils.py
gavin-aguiar: done

hallvictoria (approved):

PR #1896, opened 2026-07-30
build: migrate to CFS

hallvictoria:
## Description
Onboards all Python worker pipeline dependency restores to the Azure Artifacts public/upstream-public feed, eliminating direct traffic to PyPI and api.nuget.org.

### Changes
**pip / uv (Python packages)**
- All `PipAuthenticate@1` steps now use `public/upstream-public` with `onlyAddExtraIndex: false`, routing pip restores through the CFS
- Shell scripts forward `PIP_INDEX_URL` (set by `PipAuthenticate`) to `UV_DEFAULT_INDEX` so `uv` respects the same feed
- Removed `--prerelease=allow` from the main `uv` install command, replacing it with `--prerelease=if-necessary-or-explicit` to prevent transitive dependencies (e.g. aiohttp 4.0.0a1) from resolving to pre-release versions unnecessarily
- `pip-audit` removed from all vulnerability scan steps

**NuGet (.NET packages)**
- `nuget.config` updated to use `upstream-public` as the single package source
- Removed the per-build `"Configure NuGet feed"` bash steps that conditionally sed-patched `nuget.config` at runtime
- `ArtifactFeed` and `NuGetServiceConnection` parameters removed from job templates and all call sites

Fixes #

---

### Host-Worker Contract
   - If no, please explain why:   

### Worker Execution Logic

**Python Version Coverage**
   - If no, please explain why:

### DIFF
--- a/eng/templates/official/jobs/ci-core-tools-tests.yml
+++ b/eng/templates/official/jobs/ci-core-tools-tests.yml
@@ -13,6 +13,8 @@ jobs:
       inputs:
         artifactFeeds: 'public/upstream-public'
         onlyAddExtraIndex: false
+    - task: NuGetAuthenticate@1
+      displayName: 'Authenticate NuGet to CFS'
     - task: UsePythonVersion@0
       displayName: 'Install Python'
       inputs:
--- a/eng/templates/official/jobs/ci-core-tools-tests.yml
+++ b/eng/templates/official/jobs/ci-core-tools-tests.yml
@@ -11,7 +11,7 @@ jobs:
     - task: PipAuthenticate@1
       displayName: 'Pip Authenticate'
       inputs:
-        artifactFeeds: 'internal/upstream'
+        artifactFeeds: 'internal/PythonWorker_Internal_PublicPackages'
         onlyAddExtraIndex: false
     - task: UsePythonVersion@0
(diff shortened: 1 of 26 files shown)


gavin-aguiar (dismissed):

gavin-aguiar (dismissed):

eng/templates/official/jobs/ci-core-tools-tests.yml
```diff
@@ -9,9 +9,10 @@ jobs:
     
     steps:
     - task: PipAuthenticate@1
-      displayName: 'Pip Authenticate'
+      displayName: 'Authenticate pip to CFS'
```
fabiocav, line 12: The test extensions are built with NuGet, but this job only authenticates pip. Can you please add `NuGetAuthenticate@1` before invoking the script? Anonymous restores can consume packages already saved in the public feed, but an authenticated identity with upstream permissions is required when a package or version has not yet been ingested.

This also makes the job consistent with the other NuGet jobs and compatible with future authentication enforcement.

hallvictoria: Updated & added `NugetAuthenticate` for this pipeline

gavin-aguiar (approved):

